PR #4, opened 2015-06-25
Add /customer-notifications endpoint and definition.

caseybessette:
This pull request is for: https://github.com/azurestandard/beehive/issues/1007

### DIFF
--- a/public.json
+++ b/public.json
@@ -2161,6 +2161,14 @@
             "description": "only return (in)active notifications",
             "required": false,
             "type": "boolean"
+          },
+          {
+            "name": "person",
+            "in": "query",
+            "description": "person ID to filter by (exclude notifications dismissed by this person)",
+            "required": false,
+            "type": "integer",
+            "format": "int64"
           }
         ],
         "responses": {
--- a/public.json
+++ b/public.json
@@ -2219,7 +2219,7 @@
     },
     "/notification-dismissals": {
       "post": {
-        "summary": "Create a notification dismissal",
+        "summary": "Create a notification dismissal (a user may only dismiss notifications for himself)",
         "operationId": "addNotificationDismissal",
         "tags": [
           "notification"
--- a/public.json
+++ b/public.json
@@ -2197,6 +2197,40 @@
           }
         }
       }
+    },
+    "/notification-dismissals": {
+      "post": {
+        "summary": "Create a notification dismissal",
+        "operationId": "addNotificationDismissal",
+        "tags": [
+          "notification-dismissal"
+        ],
+        "parameters": [
+          {
+            "name": "notification-dismissal",
+            "in": "body",
+            "description": "Notification dismissal to add",
+            "required": true,
+            "schema": {
+              "$ref": "#/definitions/notificationDismissal"
+            }
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "notification dismissal response",
+            "schema": {
+              "$ref": "#/definitions/notificationDismissal"
+            }
+          },
+          "default": {
+            "description": "unexpected error",
+            "schema": {
+              "$ref": "#/definitions/errorModel"
+            }
+          }
+        }
+      }
     }
   },
   "definitions": {
@@ -2995,6 +3029,28 @@
       "required": [
         "id",
         "message"
+      ]
+    },
+    "notificationDismissal": {
+      "description": "a dismissal used to hide a read notification",
+      "type": "object",
+      "properties": {
+        "id": {
+          "type": "integer",
+          "format": "int64"
+        },
+        "notification": {
+          "type": "integer",
+          "format": "int64"
+        },
+        "person": {
+          "type": "integer",
+          "format": "int64"
+        }
+      },
+      "required": [
+        "notification",
+        "person"
       ]
     },
     "errorModel": {
--- a/public.json
+++ b/public.json
@@ -2147,29 +2147,29 @@
         }
       }
     },
-    "/customer-notifications": {
+    "/notifications": {
       "get": {
-        "summary": "Returns all customer notifications from the system that the user has access to",
-        "operationId": "findCustomerNotifications",
+        "summary": "Returns all notifications from the system that the user has access to",
+        "operationId": "findNotifications",
         "tags": [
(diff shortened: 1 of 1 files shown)


public.json
```diff
@@ -2146,6 +2146,91 @@
...
+            "name": "active",
+            "in": "query",
+            "description": "only return (in)active notifications",
+            "required": false,
+            "type": "boolean"
+          },
+          {
+            "name": "person",
+            "in": "query",
+            "description": "person ID to filter by (exclude notifications dismissed by this person)",
```
wking: Hmm, I'd been thinking of this as “some notifications may only target a subset of customers”.  For example, a “your truck is late” notification will only be associated with customers who still need to meet that truck to get their order.  But dissmissed-by-ness makes sense too.  How does that interact with activeness (@tompetersjr, this is more of a business-logic question, so maybe you can weigh in)?  I'd guess activeness lets employees collaborate on wording before publishing a particular notification.  In that case, we end up in the many-to-* squashing situation that we had with favorites (many people may favorite a product, so how do we represent that in the product-filtering API?).  Maybe follow 2ab804370 and have the following parameters:

* active, boolean, only return (in)active notfications.
* filter-person, array of integers, only include results associated with this person or people
* annotate-person, integer, required with dismissed.  Add a boolean dismissed field to the results with whether or not the annotate-person has dismissed this notification (unset for notifications that aren't associated with the annotate-person)
* dissmissed, boolean, only return notifications that are (not) dismissed by the annotate-person

so customers (who don't have the permission to see inactive notifications or see other's notifications) could use `?annotate-person={their-id}&dissmissed=false`, but someone with more credentials would want to use the more explicit `?filter-person={their-id}&annotate-person={their-id}&active=true&dissmissed=false`.

public.json
```diff
@@ -2146,6 +2146,91 @@
...
+          }
+        }
+      }
+    },
+    "/notification-dismissals": {
+      "post": {
+        "summary": "Create a notification dismissal",
+        "operationId": "addNotificationDismissal",
+        "tags": [
+          "notification-dismissal"
```
wking: This is only a separate endpoint to avoid possible collisions with a possible future `DELETE /notification/{id}` that means “actually delete from the database”, so:

* The URL should probably be: `POST /notification/{id}/dismiss`, following the (unspec'ed, shame on me) `/packaged-product/{code}/category/{cat-id}` URLs.  That keeps “do something to `/notification/{id}` that deals with that particular notification but reaches outside the model itself in a similar namespace to actions on the model itself (in this case, we're reaching outside `CustomerNotification` and into `CustomerNotificationDismissal`).

* The tag should be `notification`, so it gets grouped with the other notification endpoints.  Nobody would be using this endpoint without also being close to understanding the rest of the notification endpoints, so splitting it out into its own tag just makes it less discoverable.

public.json
```diff
@@ -2146,6 +2146,91 @@
...
+        ],
+        "parameters": [
+          {
+            "name": "notification-dismissal",
+            "in": "body",
+            "description": "Notification dismissal to add",
+            "required": true,
+            "schema": {
+              "$ref": "#/definitions/notificationDismissal"
```
wking: I'd just return the dismissed notification here (with the annotate-person being the person for which the notification was dismissed).

wking: Oops, that^ commend should have been about the 200 response payload.

public.json
```diff
@@ -2926,6 +3011,48 @@
...
+        "id": {
+          "type": "integer",
+          "format": "int64"
+        },
+        "message": {
+          "type": "string"
+        },
+        "active": {
+          "type": "boolean"
+        }
```
wking: Also `dismissed`, boolean, which is only set when we have an annotate-person.

public.json
wking: Actually, scratch my comment here.  This is fine how it stands in this PR.

public.json
wking: Scratch these comments.  It's fine how it stands.

wking:
Sorry for the shifty positions.  I think [these][1] [suggestions][2] are good, because they let you do an implicit join to only get (not) dismissed notifications for a particular annotate-person without having to make requests to both a notifications endpoint and a dismissal endpoint.  However, after some initial skepticism, having a separate `/notification-dismissals` endpoint namespace for directly manipulating dismissals makes more sense than the hacked category association example I was talking about [here][3] (and retracted [here][4]).  Ping me if I haven't clarified my current feelings ;).

[1]: https://github.com/azurestandard/api-spec/pull/4#discussion_r33535567
[2]: https://github.com/azurestandard/api-spec/pull/4#discussion_r33536852
[3]: https://github.com/azurestandard/api-spec/pull/4#discussion_r33535822
[4]: https://github.com/azurestandard/api-spec/pull/4#discussion_r33536931

caseybessette:
@wking Trevor, perhaps it would be a good idea to discuss this in a Hangout when you have some time.  I was designing the API endpoints to interact with customer notifications as they *currently* exist in the database.  Their design is definitely geared towards the current use case: draft-mode/published-mode dismissable announcements for the customer website.  They aren't even "notifications" in the way we normally use that word, as a more ephemeral "heads up" that something has happened and you might want to take "note" of it (you have a new email, someone created a pull request, you've gotten a new text message).  Probably a more appropriate term for what we call a "customer notification" is an "announcement."  These are announcements that are somewhat important for customers to see, so important that they persist until the customer explicitly says "go away, I recognize you."  I dont think that it's going to be pretty trying to combine the ephemeral "notification" concept with the "announcement" concept at a later date.  Perhaps it would be best to clarify this distinction now, then solidify that distinction in the way the API is spec-ed, and make an attempt at a later date to rework Beehive and the database to conform to the new naming.

@tompetersjr Thoughts?

caseybessette:
If we are also going to be creating a vendor portal, we may want to make vendor announcements, too.  So these announcements are really "website announcements" and they could conceivably be directed towards customers (via the website), vendors (via the vendor portal), and staff (via Beehive).  They have a draft mode (`active`) and can be dismissed.  But I think this is quite distinct from the way we may want to implement a notification system in the future.

caseybessette:
Merged to master after squashing and rebasing.

PR #7, opened 2015-07-28
Add support for POSTing to /person and /person/{id}.

### LINKED ISSUE #5: Person API: Implement POSTing.
To go with https://github.com/azurestandard/beehive/issues/1065.

caseybessette:
This allows users to change their names, email addresses, and passwords.

Fixes #5.

### DIFF
--- a/public.json
+++ b/public.json
@@ -2106,6 +2106,53 @@
             }
           }
         }
+      },
+      "put": {
+        "summary": "Update an existing person",
+        "operationId": "updatePerson",
+        "tags": [
+          "person"
+        ],
+        "parameters": [
+          {
+            "name": "id",
+            "in": "path",
+            "description": "ID of person to update",
+            "required": true,
+            "type": "integer",
+            "format": "int64"
+          },
+          {
+            "name": "person",
+            "in": "body",
+            "description": "Person to update",
+            "required": true,
+            "schema": {
+              "$ref": "#/definitions/updatePerson"
+            }
+          },
+          {
+            "name": "user-password",
+            "in": "body",
+            "description": "The authenticated person's password, required when changing person.password or person.email",
+            "required": false,
+            "type": "string"
+          }
+        ],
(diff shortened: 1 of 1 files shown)


public.json
```diff
@@ -2054,6 +2054,58 @@
...
         }
+      },
+      "post": {
+        "summary": "Update a person by ID",
+        "operationId": "updatePersonById",
+        "tags": [
+          "person"
+        ],
+        "parameters": [
+          {
```
wking: You need the `id` parameter for the path in this array.  See `/person/{id}'.get.parameters` for an example.

wking:
To match the rest of the API, we should be using PUT instead of POST
for /person/{id}.  From [1]:

  The PUT method requests that the enclosed entity be stored under the
  supplied Request-URI.

And to avoid splitting the person schema between the parameters here
and the spec in definitions.person, I'd replace all of your PUT body
parameters (except user-password) with a single $ref parameter.  For
silly examples, see the current updateDrop and updateRoute.  For a bit
more meat, see the spec around newPaymentMethod and paymentMethod.
For previous comments about why I'm cheating for updateDrop and
updateRoute, see 03e139a0, 168a97b4, 1616442d, and 27ba2a1c.  In the
route and drop cases, I've slightly broke the spec (by including
nominally requiring ‘id’).  In this case, we have the push-only
password field as well, so I'd follow the paymentMethod example and
add an updatePerson definition with the spec for the pushed fields.

[1]: http://tools.ietf.org/html/rfc2616#section-9.6

caseybessette:
Added the `id` parameter for `PUT /person/{id}`, referenced the [yet-to-be-merged](https://github.com/azurestandard/api-spec/pull/9) `updatePerson` definition instead of spelling it all out each time, and switched out `POST` for `PUT`.

**Note:** This commit depends on https://github.com/azurestandard/api-spec/pull/9 being merged.

PR #9, opened 2015-08-04
public.json: Add updatePerson

wking:
We collect several person-related attributes when registering new
people (usually customers).  Instead of duplicating schema between
parameters in POST /registration/register and the coming PUT
/person/{id}, collect that new/updated-person schema in a definition
and use *that* in the pushed body.

For silly examples of this approach, see the current updateDrop and
updateRoute. For a bit more meat, see the spec around newPaymentMethod
and paymentMethod. For previous comments about why I'm cheating for
updateDrop and updateRoute, see 03e139a, 168a97b, 1616442, and
27ba2a1. In the route and drop cases, I've slightly broke the spec (by
including nominally requiring 'id').

Related to #7, azurestandard/azure-angular-providers#2,
azurestandard/beehive#1089, and azurestandard/beehive#1025.

### DIFF
--- a/public.json
+++ b/public.json
@@ -2243,38 +2243,19 @@
             "type": "string"
           },
           {
-            "name": "name",
+            "name": "person",
             "in": "body",
-            "description": "The registrant's full name",
+            "description": "The registrant's personal information.",
             "required": true,
-            "type": "string"
-          },
-          {
-            "name": "password",
-            "in": "body",
-            "description": "The registrant's password",
-            "required": true,
-            "type": "string"
-          },
-          {
-            "name": "email",
-            "in": "body",
-            "description": "The registrant's email address",
-            "required": true,
-            "type": "string",
-            "format": "email"
-          },
-          {
-            "name": "can-email",
-            "in": "body",
-            "description": "Set to false if you do not wish to be contacted via email after completing registration (assumed true)",
-            "required": false,
-            "type": "string"
+            "type": "object",
+            "schema": {
+              "$ref": "#/definitions/updatePerson"
+            }
           },
           {
             "name": "address",
             "in": "body",
-            "description": "The registrant's mailing address (required if the've set 'catalog' or unset 'can-email')",
+            "description": "The registrant's mailing address (required if they've set person.catalog).",
             "required": false,
             "type": "object",
             "schema": {
@@ -2297,13 +2278,6 @@
             "description": "The registrant's chosen drop ID (if any)",
             "required": false,
             "type": "string"
-          },
-          {
-            "name": "catalog",
-            "in": "body",
-            "description": "Set to true to receive the product catalog",
-            "required": false,
-            "type": "boolean"
           }
         ],
         "responses": {
@@ -3249,12 +3223,53 @@
         "email": {
           "type": "string",
           "format": "email"
+        },
+        "can-email": {
+          "description": "Can Azure contact this person for reasons other than registration",
+          "type": "boolean"
+        },
+        "catalog": {
+          "description": "Set to true to receive the product catalog",
+          "type": "boolean"
+        },
+        "company": {
+          "description": "Employer name",
+          "type": "string"
         }
       },
       "required": [
         "id",
         "name"
       ]
+    },
+    "updatePerson": {
+      "description": "a person (customer, vendor, employee, ...)",
+      "type": "object",
+      "properties": {
+        "name": {
+          "type": "string"
+        },
+        "password": {
+          "description": "Password for authentication",
+          "type": "string"
+        },
+        "email": {
+          "type": "string",
+          "format": "email"
+        },
+        "can-email": {
+          "description": "Can Azure contact this person for reasons other than registration",
+          "type": "boolean"
+        },
+        "catalog": {
+          "description": "Subscribed to receive the product catalog",
+          "type": "boolean"
+        },
+        "company": {
+          "description": "Employer name",
+          "type": "string"
+        }
+      }
     },
     "address": {
       "description": "an address following the convention of http://microformats.org/wiki/h-adr and https://tools.ietf.org/html/rfc6350#section-6.3.1, with an additional \"name\" property",


wking:
Ping @mercenaryjosh, @caseybessette.  I think this helps make the spec DRYer.  But you both have worked on the implementation-side of the related parts of this spec, so I'm interested to see how this reorganization looks to you.

MercenaryJosh:
The /registration/register spec is missing the fields company, password, customer_type and contact_method. 

The company field is an optional field required by the website spec.

The password field is required in order to register a user.

The customer_type field is used for data verification. If the customer type is 'drop', the drop field is required. If the customer type is 'ups', a postal_code is required.

contact_method: According to the website spec, contact_method should have the options of email, phone, and text. This new api-spec contains a boolean field for can-email in the updatePerson model, but that does not reflect the customer's preferred way to be contacted. For example, a customer may be able to be emailed, but prefers to be called. The can-email field will be true, but there is nowhere to register their preference to be called. Also, since on registration we email them a confirmation, would that indicate that can-email must be true by default for registering customers?

Also, on the address model it lists locality, region, and country-name as required. In the case of a user registering as a UPS customer, the postal-code is required but no other address data is. So in this case we would have an address object without the locality, region, and country-name, but with the postal-code.

wking:
On Fri, Aug 07, 2015 at 03:07:58PM -0700, MercenaryJosh wrote:
> The /registration/register spec is missing the fields company,
> password, customer_type and contact_method.

Ah, thanks.  company, and password should go into updatePerson, and
I've updated this PR accordingly.

> The customer_type field is used for data verification. If the
> customer type is 'drop', the drop field is required. If the customer
> type is 'ups', a postal_code is required.

Can't we just handle that depending on whether or not 'drop' is
present in the registerPerson endpoint?

> contact_method: According to the website spec, contact_method should
> have the options of email, phone, and text. This new api-spec
> contains a boolean field for can-email in the updatePerson model,
> but that does not reflect the customer's preferred way to be
> contacted. For example, a customer may be able to be emailed, but
> prefers to be called. The can-email field will be true, but there is
> nowhere to register their preference to be called.

Hmm.  The telephone type has "pref", after which I think it's up to
the contractor to pick their favorite among the contactee's preferred
numbers.  We might want a separate knob for email vs. phone, but
honestly, if can-email is set I think we prefer sending email enough
to not let folks pick "prefer-phone".  Folks who really like phone
more can just turn off can-email.  @tompetersjr, want to weigh in?
Maybe we can punt here, and add a separate phone-vs.-email choice
field if we get too much pushback from overloading can-email.

> Also, since on registration we email them a confirmation, would that
> indicate that can-email must be true by default for registering
> customers?

That's why there's the registration special case:

  "can-email": {
    "description": "Can Azure contact this person for reasons other than registration",
    "type": "boolean"
  },

> Also, on the address model it lists locality, region, and
> country-name as required. In the case of a user registering as a UPS
> customer, the postal-code is required but no other address data is.

I think we should still require country-name, because otherwise
postal-code isn't very meaningful.  I've taken a stab at getting this
clarified in the just-pushed version of this PR (assuming
postal-code/country addresses are acceptable), but note that we use
primary addresses for other purposes, and we'll want a mailable
address for most of those [1].  In any case, I guess we'd collect the
rest of this address information from UPS customers if/when they
actually place an order.  I'm not sure if/when we'd get address
information from drop customers, but for most new folks (who won't
have COD terms) we'd at least have a credit-card billing address we
could use for dunning and the like.  @tompetersjr, do you know how
widely postal-code/country addresses will be acceptable vs. mailable
addresses?

[1]: https://docs.google.com/document/d/1dwayyitTQT5ma-UhFfkC16i6aVrJ2ArV-HnUdWyst10/edit#heading=h.yigcrdyq7qzm

wking:
On Fri, Aug 07, 2015 at 03:57:10PM -0700, W. Trevor King wrote:
> Fri, Aug 07, 2015 at 03:07:58PM -0700, MercenaryJosh:
> > The customer_type field is used for data verification. If the
> > customer type is 'drop', the drop field is required. If the
> > customer type is 'ups', a postal_code is required.
> 
> Can't we just handle that depending on whether or not 'drop' is
> present in the registerPerson endpoint?

Wrinkling this even more, some folks are going to be using the
registration endpoint to register vendors, and they may be neither
drop nor UPS customers.  I think the current:

  The registrant's mailing address (required if they've set
  person.catalog or unset person.can-email).  If can-email is true,
  only the postal code and country name are required (and not the
  other usually-required address properties).

still covers that case, but I'm not clear on how vendor contact
usually works.

MercenaryJosh:
The third option for customer_type  when a new customer is registering through the website is that they will be a 'will call' customer. This would mean that both drop and postal-code would not be required. This is similar to registering vendors that are neither drop nor UPS customers.

wking:
On Fri, Aug 07, 2015 at 04:30:09PM -0700, MercenaryJosh wrote
> The third option for customer_type when a new customer is
> registering through the website is that they will be a 'will call'
> customer. This would mean that both drop and postal-code would not
> be required. This is similar to registering vendors that are neither
> drop nor UPS customers.

I'm not sure if we're planning on addressing will-call explicitly in
the API, but in Beehive we currently have will-call drops [1,2] (on
routes W and WD, respectively).  So I think those folks will still be
setting ‘drop’ in their registration request.

I actually think using drops for the will-call locations makes a lot
of sense, so I don't see a need for any explicit will-call handling.
That's just me though, not part of an official spec.

[1]: https://beehive.azurestandard.com/drop/109/detail
[2]: https://beehive.azurestandard.com/drop/111/detail

MercenaryJosh:
According to @tompetersjr comment on the website spec[1], will-call will always be at the warehouse in Moro. So if the customer registers as a will-call customer, should drop D111 be set as their drop? Or perhaps give them a choice between D109 and D111 (Durfur and Moro, respectively) when they choose will-call?

If the D109 or D111 drops are used for will-call customers than we could validate that if there is no drop, then there must be a postal-code for customer registration. The customer_type field (or renamed as appropriate) may not be needed, but could be valuable for letting the api know what type of user is being registered and allowing the different types of users (customer, vendor, etc.) to have different required data fields.

[1] https://docs.google.com/document/d/1dwayyitTQT5ma-UhFfkC16i6aVrJ2ArV-HnUdWyst10/edit#heading=h.knkm0ikdua62

wking:
On Fri, Aug 07, 2015 at 06:12:06PM -0700, MercenaryJosh wrote:
> According to @tompetersjr comment on the website spec[1], will-call
> will always be at the warehouse in Moro. So if the customer
> registers as a will-call customer, should drop D111 be set as their
> drop? Or perhaps give them a choice between D109 and D111 (Durfur
> and Moro, respectively) when they choose will-call?

Choice sounds like the best plan to me.  I've left a comment to that
effect in the spec as well, so we can keep these conversations
connected.

> If the D109 or D111 drops are used for will-call customers than we
> could validate that if there is no drop, then there must be a
> postal-code for customer registration.

That's closer to the current Google Docs wording, but as I discuss in
the f5c37bb commit message, I think that spec is too strict for drop
customers, and I prefer my proposed can-email switching.

> The customer_type field (or renamed as appropriate) may not be
> needed, but could be valuable for letting the api know what type of
> user is being registered and allowing the different types of users
> (customer, vendor, etc.) to have different required data fields.

If we need different validation, then we definitely need a field(s) to
select which validation to use.  But I have a hard time imagining that
that *validation* is the only thing that would need to change.  If
possible, I'd rather attach the validation to whatever relationship
actually requires it (e.g. if you ask for a catalog, we need to know
where to (e)mail it).  The validation is still in flux here, so no
promises, but I'd expect we can figure it out without resorting to a
validation-only switch.

MercenaryJosh:
I agree that if the customer elects to receive a catalog, not every field should be required as is implied in the spec. I would think that the normal address fields that would be needed to fulfill the request should be required (street-address, locality, postal-code, etc.).

public.json
```diff
@@ -2191,38 +2191,19 @@
...
+            "type": "object",
+            "schema": {
+              "$ref": "#/definitions/updatePerson"
+            }
           },
           {
             "name": "address",
             "in": "body",
-            "description": "The registrant's mailing address (required if the've set 'catalog' or unset 'can-email')",
+            "description": "The registrant's mailing address (required if they've set person.catalog or unset person.can-email).  If can-email is true, only the postal code and country name are required (and not the other usually-required address properties)."
```
caseybessette: You need a comma at the end of this line.

caseybessette:
This looks good to me.

public.json
wking: On Mon, Aug 10, 2015 at 08:43:14AM -0700, Casey M. Bessette wrote:
> > -            "description": "The registrant's mailing address (required if the've set 'catalog' or unset 'can-email')",
> > +            "description": "The registrant's mailing address (required if they've set person.catalog or unset person.can-email).  If can-email is true, only the postal code and country name are required (and not the other usually-required address properties)."
> 
> You need a comma at the end of this line.

Thanks, fixed in the just-pushed 751c5b3.

PR #12, opened 2015-08-11
Add GET /packaged-product-tags to provide all tags.

caseybessette:


wking:
On Tue, Aug 11, 2015 at 01:28:41PM -0700, Casey M. Bessette wrote:
>   * Add GET /packaged-product-tags to provide all tags.

This looks good to me, but if packagedProductTag is just going to be
an alias for string, I'd rather drop it from the definitions and have:

  "description": "packaged product tag response",
  "schema": {
    "type": "array",
    "items": {
      "type": "string"
    }
  }

caseybessette:
Removed `packagedProductTag` definition and just used `"type": "string"` where needed.

caseybessette:
@wking Actually, are we going to want human readable names along with this?  We're going to want the `no-gmo` tag to display as `No GMO` for users.  Maybe this should be returning both a `tag` and a `name` -- the tag can be unique and used as the "key".

wking:
On Thu, Aug 13, 2015 at 12:16:21PM -0700, Casey M. Bessette wrote:
> Actually, are we going to want human readable names along with this?
> We're going to want the `no-gmo` tag to display as `No GMO` for
> users.  Maybe this should be returning both a `tag` and a `name` --
> the tag can be unique and used as the "key".

Good catch.  With the hard-coded enum approach, we could leave it to
clients to provide their own human-readable versions of the tags
listed in the spec.  Now that the tags are more dynamic, we probably
want to restore packagedProductTag, and have a structure like:

  {
    "slug": "no-gmo",
    "name": "No GMO"
  }

Once we do *that*, it might be worth having an additional
"description" field, or a way to get a longer-form description for
tooltips, etc.  It would be nice to use the short-forms everywhere as
bare strings, return these more elaborate objects from
/packaged-product-tags, and leave it to the UI to lookup the
additional detail for a given tag entry if they're interested.

caseybessette:
Added `packagedProductTag` definition back as an object with `slug` and `name` properties.

Issue #15, opened 2015-08-28
Add API spec for setting primary address


heuscher:
See section 8.3.2 Address Book, in the website spec doc.

Customers need the ability to change their primary address in the Address Book of the new website.

tompetersjr:
This is now functioning.

PR #17, opened 2015-08-31
public.json: Add address.preference

wking:
We need a consistent way to select a user's primary address, and the
cleanest API for that seems to be letting the user order all their
preferences.  Then setting a new primary address would look something
like (using [azure-angular-providers][1]):

    newPrimary.preference = 100;
    oldPrimary.preference = 0;
    newPrimary.$save().$promise.then(function () {
      oldPrimary.$save();
    });

There's a low probability that the `newPrimary.$save` goes through and
the `oldPrimary.$save` does not.  If the `oldPrimary.preference` was
originally 100, that means there is some time (while the
`oldPrimary.$save` is in flight, and possibly after if that call fails)
where both addresses will have the same preference on the server.  If
that's a concern, you could use something like:

    newPrimary.preference = oldPrimary.preference + 1;
    oldPrimary.preference = 0;
    newPrimary.$save().$promise.then(function () {
      oldPrimary.$save().$promise.then(function () {
        newPrimary.preference -= 1;
        newPrimary.$save();
      });
    });

If you weren't worried about rolling over the int32 (you're probably
safe ;), you could skip the decrement step.

[1]: https://github.com/azurestandard/azure-angular-providers

### DIFF
--- a/public.json
+++ b/public.json
@@ -2038,7 +2038,7 @@
     "/addresses": {
       "get": {
         "summary": "Returns all addresses from the system that the user has access to",
-        "description": "Addresses associated with a person are ordered by decreasing preference.  The ordering between addresses associated with different people is undefined.",
+        "description": "Addresses associated with a person are ordered by decreasing preference.  The ordering between addresses associated with different people or (for the same person) with the same preference value is undefined.",
         "operationId": "findAddresses",
         "tags": [
           "address"
@@ -3577,6 +3577,11 @@
         },
         "country-name": {
           "type": "string"
+        },
+        "preference": {
+          "description": "ordering precedence for the findAddresses.  Defaults to zero.",
+          "type": "integer",
+          "format": "int32",
         }
       },
       "required": [


wking:
Fixes #15.

caseybessette:
So there can be more than one primary address?  This `preference` order is not going to be unique?  I'm not sure what the UI is going to look like, but is it going to be able to handle more than one primary address?  And if we ever do any automatic shipping (do we do this?), we're just going to randomly select one of the primary addresses?

wking:
On Fri, Sep 11, 2015 at 09:09:18AM -0700, Casey M. Bessette wrote:
> So there can be more than one primary address?  This `preference`
> order is not going to be unique?

I don't care if its unique, but it seems easier to not require it.  In
the event of a tie, the order is undefined (left up to the
implementation and possibly not stable, although I expect we would be
stable).

> I'm not sure what the UI is going to look like, but is it going to
> be able to handle more than one primary address?

You could have a drag & drop UI for ordering the addresses, but I
think we can just have a “make primary” button (details in the commit
message and initial post).

> And if we ever do any automatic shipping (do we do this?), we're
> just going to randomly select one of the primary addresses?

Yes, and yes.  For stuff that gets mailed to the primary address
without user intervention, see the Google Doc spec.

caseybessette:
Great, thanks for the explanation. :)

PR #19, opened 2015-09-04
public.json: Add 'GET /registration/confirm?token=...'

wking:
This allows us to support the registrant copy/pasting confirmation
URLs into their browser (it's harder to get your browser to POST
without an HTML form or such).

### DIFF
--- a/public.json
+++ b/public.json
@@ -2460,6 +2460,36 @@
       }
     },
     "/registration/confirm": {
+      "get": {
+        "summary": "Complete an in-progress registration",
+        "operationId": "confirmRegistrationGet",
+        "tags": [
+          "registration"
+        ],
+        "parameters": [
+          {
+            "name": "token",
+            "in": "query",
+            "description": "The registrant's confirmation token",
+            "required": true,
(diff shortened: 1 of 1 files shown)


public.json
```diff
@@ -2460,6 +2460,36 @@
       }
     },
     "/registration/confirm": {
+      "get": {
+        "summary": "Complete an in-progress registration",
+        "operationId": "confirmRegistration",
```
caseybessette: The `operationId` needs to be unique according to the [Swagger spec](http://swagger.io/specification/):
> Unique string used to identify the operation. The id MUST be unique among all operations described in the API. Tools and libraries MAY use the operationId to uniquely identify an operation, therefore, it is recommended to follow common programming naming conventions.

wking: On Fri, Sep 11, 2015 at 09:13:39AM -0700, Casey M. Bessette wrote:
> > +        "operationId": "confirmRegistration",
> 
> The `operationId` needs to be unique…

Ah, thanks.  Should be fixed with 47ec31b → 182d6c9.

PR #21, opened 2015-09-08
Add category property to packagedProductTag definition.

caseybessette:
This will be used by the frontend to filter unwanted tags from being
displayed.   This way product status tags, e.g., can be distinguished
from other tag types.

### DIFF
--- a/public.json
+++ b/public.json
@@ -3478,6 +3478,15 @@
         "name": {
           "description": "human-readable tag name",
           "type": "string"
+        },
+        "category": {
+          "description": "category code describing the type of tag",
+          "type": "string",
+          "enum": [
+            "status",
+            "price-level",
+            "characteristic"
+          ]
         }
       }
     },


wking:
On Tue, Sep 08, 2015 at 01:27:58PM -0700, Casey M. Bessette wrote:
> This will be used by the frontend to filter unwanted tags from being
> displayed.

If tags aren't going to be a flat space, it probably makes sense to
handle them in the same way we currently handle the category tree.

caseybessette:
I spoke with Trevor about this and we agreed to drop it.

tompetersjr:
Can you expand on the reasoning?  Also how will we distinguish between a "My Filters" tag and the search filters for "New, On Sale, Overstock and Bargain Bin"?

wking:
On Tue, Sep 08, 2015 at 03:58:35PM -0700, Tom Peters wrote:
> Can you expand on the reasoning?

It was mostly that adding categories makes the tag system pretty much
the same as the existing category system.  We don't want to use the
Category tooling for PieceMeta.characteristic_set (see 9fa548247,
apps/warehouse_pieces/models.py: Shift Characteristic from Ingredients
to PieceMetas, 2015-05-13), and information like 'on-sale-retail' is
coming from a non-tree structure anyway, so it seemed to awkward to
build in the full category-style handling, and I don't want to lock in
the single categorization level that was floated in this PR.

> Also how will we distinguish between a "My Filters" tag and the
> search filters for "New, On Sale, Overstock and Bargain Bin"?

I was just going to put all of those into “My Filters”, since they're
boolean switches for filtering search results (just like the other “My
Filters” tags).  Is there a reason not to use a single UI for all
tags?  If we want a special UI for those tags (e.g. show them in the
upper right, instead of along the left sidebar), that seems easy
enough to code into azurestandard/website without adjusting the API.

caseybessette:
Reopening per decision in https://github.com/azurestandard/beehive/issues/1345

caseybessette:
Rebased off master.

public.json
```diff
@@ -3478,6 +3478,10 @@
         "name": {
           "description": "human-readable tag name",
           "type": "string"
+        },
+        "category": {
+          "description": "category code describing the type of tag",
+          "type": "string"
```
wking: I'd rather make this an enum (listing [these][1] slugs).  #12 landed as a utility so clients didn't have to hard-code the list of tags (which we didn't specify in the API spec).  If this lands, they'll need some way to get a list of tag categories.  I expect that list to be small and stable enough to hard-code it in the enum here (from which client authors can hard-code it in their clients).  Adding a `/packaged-product-tag-categories` endpoint or making `/packaged-product-tags` support a tree (like `/categories`) both seem like they're going too far ;).

[1]: https://github.com/azurestandard/beehive/pull/1383/files#diff-e887c6cc5c21fb803e9a520fc028027cR352

caseybessette:
Updated with your enum suggestion.  Good idea, thanks.

PR #22, opened 2015-09-11
public.json: Add public=? boolean to /packaged-product-tags

wking:
A user who is authorized to view all tags may not want to view
internal tags (e.g. they're hitting this endpoint while shopping for
themselves).  This option allows UI code to explicitly set
`?public=true` when they don't intend to access those internal tags.
I expect most UIs will be setting `?public=true`, but defaulting to
limited results (and requiring folks to specify `?all=true` or some
such) would break our usual pattern of “show everything and filter
with query parameters”.

### DIFF
--- a/public.json
+++ b/public.json
@@ -1217,6 +1217,13 @@
         ],
         "parameters": [
           {
+            "name": "internal",
+            "in": "query",
+            "description": "if true, only return internal tags (that the user has access to).  If false, only return public tags (regardless of the user's access to internal tags).  If unset, don't filter based on the tag visibility",
+            "required": false,
+            "type": "boolean"
+          },
+          {
             "name": "limit",
             "in": "query",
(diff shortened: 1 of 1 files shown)


public.json
```diff
@@ -1217,6 +1217,13 @@
         ],
         "parameters": [
           {
+            "name": "public",
+            "in": "query",
+            "description": "if true, only return public tags (regardless of the user's access to internal tags).  If false, only return internal tags (that the user has access to).  If unset, don't filter based on the tag visibility"
```
caseybessette: I approve of this PR. But you might want to add a comma at the end of this line. ;)

One suggestion, though, for you to consider.  We could call this parameter `internal` and do just this opposite (require the UI to generally add `?internal=false`), thereby being consistent with the naming elsewhere (we deal with "internal tags" but haven't spoken of "public tags").

PR #26, opened 2015-09-15
public.json: Add drop.order-minimum

wking:
On Tue, Sep 15, 2015 at 12:19:21PM -0700, heuscher [wrote][1]:
> Reference Google spec document 8.7.1 Current Drop.
>
> Need to return the minimum, total order amount
> (`drop.minimum_order_amount`) required for a drop.

Part of azurestandard/beehive#1192.

[1]: https://github.com/azurestandard/beehive/issues/1192

### DIFF
--- a/public.json
+++ b/public.json
@@ -2772,6 +2772,12 @@
         "fees": {
           "$fees": "#definitions/dropFees"
         },
+        "order-minimum": {
+          "description": "the minimum total order value (in dollars) required to ship a particular trip to this drop.  For example, if the minimum is $400 and there are three $100 orders placed for a stop, that stop will be under-minimum, and the orders will not be shipped.  If, on the other hand, there were four $100 orders placed for a stop, that stop would be (just) over-minimum, and the orders would be shipped.",
+          "type": "number",
+          "format": "float"
+          }
+        },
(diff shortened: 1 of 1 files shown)


wking:
@tompetersjr, I've assigned this to you, because I think review will mostly involve vetting the wording I've used in the description.  I think my phrasing is accurate, but it may not be what we want to use in public specs.

public.json
```diff
@@ -2772,6 +2772,12 @@
         "fees": {
           "$fees": "#definitions/dropFees"
         },
+        "order-minimum": {
+          "description": "the minimum total order value (in dollars) required to ship a particular trip to this drop.  For example, if the minimum is $500 and there are three $100 orders placed for a stop, that stop will be under-minimum, and the orders will not be shipped.  If, on the other hand, there were five $100 orders placed for a stop, that stop would be (just) over-minimum, and the orders would be shipped.",
```
tompetersjr: I think this is a good description.  The only thing I might change would be the $500 to something like $400 and make the examples match that.  Just to make the example a bit different then what the exact in-house business process is.  Not sure that really matters but it wouldn't hurt anything.

PR #32, opened 2015-09-22
Favorites API spec

colin-merc-tech:


public.json
```diff
@@ -2146,6 +2146,157 @@
...
+          },
+          "default": {
+            "description": "unexpected error",
+            "schema": {
+              "$ref": "#/definitions/errorModel"
+            }
+          }
+        }
+      },
+    "/favorites/{id}": {
```
wking: The other API endpoints are all `/{plural}` for a listing GET and a creating POST, and then `/{singular}/{id}` for a reading GET, a clobbering PUT, or a DELETE.  We should continue that pattern here with `/favorite/{id}`.

public.json
```diff
@@ -2910,6 +3061,51 @@
...
+          },
+          "size": {
+            "type": "string"
+          },
+          "price": {
+            "type": "string"
+          },
+          "is_checked": {
+            "type": "boolean"
+          }
```
wking: This is a lot of per-product information to inline in the favorite model.  I'd rather just keep a reference (e.g. the packaged-product's code) and have folks hit the (packaged-)product API to get more information.  That's easy with the Beehive API's filter-person (2ab804370, public.json: Add packagedProduct.favorite boolean and annotate-person parameter, 2015-06-08), but it doesn't scale as well to the more-static Algolia indexes.  On the other hand, Algolia may be fast enough that two requests (one to get your favorites, and another to get the associated (packaged-)products) may be fine.  If we need to optimize more, I'd be ok with optionally embedding the associated descriptions/product object here as `product`, and then always embedding that information when we push to Algolia.  Does that sound like a workable plan?

public.json
```diff
@@ -2910,6 +3061,51 @@
...
+        }
+      }
+    },
+    "favorite delete": {
+      "description": "a customer favorite delete response",
```
wking: The other API endpoints just 204 deletions without a model-specific return type, so I think we should follow that pattern and drop the “favorite delete” definition.

public.json
```diff
@@ -2146,6 +2146,157 @@
           }
         }
       }
+    },
+    "/favorites": {
+      "get": {
+        "summary": "Returns a list of all customer's product favorites",
+        "operationId": "listFavorites",
+        "tags": [
+          "favorites"
```
wking: Our other tags are all singular, so probably follow that pattern with “favorite” here.

public.json
```diff
@@ -2146,6 +2146,157 @@
           }
         }
       }
+    },
+    "/favorites": {
+      "get": {
+        "summary": "Returns a list of all customer's product favorites",
+        "operationId": "listFavorites",
```
wking: Our other listing endpoints all have IDs like “findDrops”, so probably follow that pattern ([borrowed from Swagger's examples][1]) with “findFavorites”.

[1]: https://github.com/swagger-api/swagger-spec/blob/master/examples/v2.0/json/petstore-simple.json#L30

public.json
```diff
@@ -2146,6 +2146,157 @@
           }
         }
       }
+    },
+    "/favorites": {
+      "get": {
+        "summary": "Returns a list of all customer's product favorites",
```
wking: To support customer-service acting on behalf of other customers (e.g. see azurestandard/beehive#1174), the other API endpoints are usually structured to show you everything you have access to (e.g. `/drops` [returns all drops from the system that the user has access to][1]).  For some situations (e.g. addresses), that's [just going to be your addresses][2], so [unless we've given you special customer-service privileges][3], we'll [forbid unlimited access][4].  Unprivileged users will always have to [specify a `filter-person` query parameter][5], which we'll use to [filter the results][6] (it looks like I still need to update the drop API after the person → filter-person change).

Let me know if there's anything in there that doesn't make sense, but I think we should update these endpoints to match that pattern.

[1]: https://github.com/azurestandard/api-spec/blob/8b3776f198ec40c133098553e638d6b19a2e4c55/public.json#L50
[2]: https://github.com/azurestandard/beehive/blob/990af5f21ccdb30ce781258b0658ef5c6868eae5/apps/api/views/address.py#L109-L110
[3]: https://github.com/azurestandard/beehive/blob/990af5f21ccdb30ce781258b0658ef5c6868eae5/apps/api/views/address.py#L115-L116
[4]: https://github.com/azurestandard/beehive/blob/990af5f21ccdb30ce781258b0658ef5c6868eae5/apps/api/views/address.py#L120
[5]: https://github.com/azurestandard/api-spec/blob/8b3776f198ec40c133098553e638d6b19a2e4c55/public.json#L87
[6]: https://github.com/azurestandard/beehive/blob/990af5f21ccdb30ce781258b0658ef5c6868eae5/apps/api/views/address.py#L103

public.json
colin-merc-tech: I'll talk to Josh and confirm how the data is going to be used on the front end and then we can go from there. I'm guessing two algolia hits will be fine, but if for some reason it isn't, we can look into optimizing like you mentioned.

public.json
colin-merc-tech: I'll update the branch but the way this is handled feels really clumsy and I think we should revisit how it's implemented. If nothing else, I think we should be separating batch queries like that into their own views and move the majority of permissions out of the view and into decorators or leverage the ApiView class to reduce the view bloat. Maybe we can discuss further in a more appropriate forum.

wking: On Wed, Sep 23, 2015 at 06:27:43AM -0700, colin-merc-tech wrote:
> Maybe we can discuss further in a more appropriate forum.

Sure.  Do you want to sketch out your alternative approach in more
details in the Hangout where we discuss this PR?  That would give us a
concrete example to anchor the more general discussion.

colin-merc-tech: That works. I'll take a closer look after the changes are up.

jeffstrickler:
Let's talk this over when we get a hangout scheduled. That said:

"To support customer-service acting on behalf of other customers (e.g. see azurestandard/beehive#1174), the other API endpoints are usually structured to show you everything you have access to (e.g. /drops returns all drops from the system that the user has access to)."

   Why embed either permissions or the superuser aliasing into the API? Make the API you want, then use decorators or other kinds of centralized facades to control permissions and access.


"For some situations (e.g. addresses), that's just going to be your addresses, so unless we've given you special customer-service privileges, we'll forbid unlimited access. Unprivileged users will always have to specify a filter-person query parameter, which we'll use to filter the results (it looks like I still need to update the drop API after the person → filter-person change)."


 We don't have a requirement to develop an API for showing all favorites for all customers. Why make the API more complex for the 80% case, especially when that complexity is for both the consumer and producer? In this case, it's also producing an API scenario that is likely to never be used.

wking:
On Wed, Sep 23, 2015 at 01:17:13PM -0700, Jeff Strickler wrote:
> Why embed…

This is probably to involved for a comment, so I'll punt to the
Hangout.

> We don't have a requirement to develop an API for showing all
> favorites for all customers. Why make the API more complex for the
> 80% case, especially when that complexity is for both the consumer
> and producer? In this case, it's also producing an API scenario that
> is likely to never be used.

You don't have to implement the unused functionality now, you just
have to pick an API that can grow into that functionality if we decide
we need it later.  For an existing example of this, the
notification-dismissal endpoint currently requires a filter-person
query argument (see the last paragraph of [1]).

For example, if we decide later that product managers want to see all
the people who've favorited QP123, you could use the same /favorites
endpoint and set a packaged-product=QP123 query paramter).  You
*could* support that use case by adding new endpoints for each type of
filtering you need

* /favorites-by-person/{person-id}
* /favorites-by-packaged-product/{packaged-product-code}

but they'd duplicate a lot of code unless you spent the time to factor
the shared stuff out.  And they wouldn't support things like:

  /favorites?filter-person={person-id}&packaged-product?{packaged-product-code}

The “push all the entries and then give ways to filter” approach also
maps well onto less dynamic engines like Algolia (although we wouldn't
want to push all of *favorites* to Algolia without setting up an
internal-user API key).

[1]: https://github.com/azurestandard/api-spec/pull/33#issue-107825344

public.json
```diff
@@ -2146,6 +2146,154 @@
...
+            "type": "integer"
+          },
+          {
+            "name": "is_checked",
+            "in": "body",
+            "description": "Default is checked",
+            "required": false,
+            "type": "boolean"
+          }
+        ],
```
wking: Swagger [only allows one body parameter][1], so you need to shift these into a `updateFavorite` definition.  See `updatePerson` in master andd1518f05  for an example.

[1]: https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#fixed-fields-4

public.json
```diff
@@ -2146,6 +2146,154 @@
...
+            "description": "Default quantity number",
+            "required": false,
+            "type": "integer"
+          },
+          {
+            "name": "is_checked",
+            "in": "body",
+            "description": "Default is checked",
+            "required": false,
+            "type": "boolean"
```
wking: Recycle the `updateFavorite` definition here.

public.json
```diff
@@ -2146,6 +2146,154 @@
...
+          },
+          {
+            "name": "quantity",
+            "in": "body",
+            "description": "Default quantity number",
+            "required": false,
+            "type": "integer"
+          },
+          {
+            "name": "is_checked",
```
wking: Drop `is_checked`, since we're not sure what it's for.  We can always add it back in later if it turns out to be important.

public.json
```diff
@@ -2910,6 +3058,26 @@
...
+      "properties": {
+        "id": {
+          "type": "integer",
+          "format": "int64"
+        },
+        "quantity": {
+          "type": "integer",
+          "format": "int64"
+        },
+        "product_code": {
```
wking: We try to match the property name to the referenced model, so probably `packaged-product` here.

public.json
```diff
@@ -2910,6 +3059,37 @@
...
+        },
+        "quantity": {
+          "type": "integer",
+          "format": "int64"
+        },
+        "packaged-product": {
+          "type": "string"
+        }
+      },
+      "updateFavorite": {
```
wking: I think this is one level too deep (it should be its own definition, not part of the `favorite` definition).

public.json
```diff
@@ -2146,6 +2146,155 @@
...
+            "description": "unexpected error",
+            "schema": {
+              "$ref": "#/definitions/errorModel"
+            }
+          }
+        }
+      },
+      "post": {
+        "summary": "Creates and returns a customer favorite",
+        "operationId": "createFavorite",
```
wking: Existing "post" operation IDs are all “addModel” ;).  E.g. see [here][1].

[1]: https://github.com/azurestandard/api-spec/blob/65084cc35839c4cc4fe3a4b5fdfcd98926eac5ca/public.json#L143

public.json
```diff
@@ -2146,6 +2146,155 @@
...
+        "tags": [
+          "favorite"
+        ],
+        "parameters": [
+          {
+            "name": "favorite",
+            "in": "body",
+            "description": "Favorite to update",
```
wking: “update” → “add” here, since we're creating a new entry.

public.json
```diff
@@ -2146,6 +2146,155 @@
...
+            "required": true,
+            "schema": {
+              "$ref": "#/definitions/updateFavorite"
+            }
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "array of favorites response",
```
wking: This isn't an array ;).  With the addDrop endpoint, I just had “drop response”, so maybe “favorite response”?

public.json
```diff
@@ -2146,6 +2146,155 @@
...
+            "schema": {
+              "$ref": "#/definitions/errorModel"
+            }
+          }
+        }
+      },
+    "/favorite/{id}": {
+      "get": {
+        "summary": "Returns a single customer favorite",
+        "operationId": "getFavorite",
```
wking: The existing API uses `find<Model>ById” for `/<model>/{id}`, so this should be `findFavoriteById`.

public.json
```diff
@@ -2146,6 +2146,155 @@
...
+        "tags": [
+          "favorite"
+        ],
+        "parameters": [
+          {
+            "name": "id",
+            "in": "path",
+            "description": "ID of favorite to update",
```
wking: GET, so not updating.  Description for findDropById's `id` parameter is “ID of drop to fetch”, so maybe “ID of favorite to fetch”.

public.json
wking: Probably add `required` blocks to these definitions too.  `favorite` should require all three properties.  `updateFavorite` should require both properties.

[0]: https://github.com/azurestandard/beehive/blob/872db47d49e25bd64ce441b0adac3fac356916db/apps/api/views/favorites.py#L40
[1]: https://github.com/azurestandard/beehive/blob/872db47d49e25bd64ce441b0adac3fac356916db/apps/api/views/favorites.py#L44
[2]: https://github.com/azurestandard/beehive/blob/872db47d49e25bd64ce441b0adac3fac356916db/apps/api/views/favorites.py#L109-L110

public.json
```diff
@@ -2703,6 +2703,177 @@
...
+        ],
+        "parameters": [
+          {
+            "name": "filter-person",
+            "in": "query",
+            "description": "person ID to create favorite for",
+            "required": false,
+            "type": "integer",
+            "format": "int64"
+          },
```
wking: Now that you have `favorite.customer` in your model, there's no need for a query parameter like this to say who's favorite this is.

PR #34, opened 2015-09-30
Add new fields to route reponse schema

colin-merc-tech:
Let me know if there's anything that doesn't look right. The fields were requested for UI work but there was no real spec so the changes could be slightly off.

code change branch: https://github.com/azurestandard/beehive/pull/1231

### DIFF
--- a/public.json
+++ b/public.json
@@ -2858,12 +2858,25 @@
           "description": "number of days after cutoff before the first stop",
           "type": "integer",
           "format": "int32"
+        },
+        "internal": {
+          "description": "Is the route for internal use? This controls whether customers can access the route via the API (they can't if internal is true).",
+          "type": "boolean"
+        },
+        "backhaul-cost-per-pound": {
+          "description": "route backhaul cost per pound",
+          "type": "string"
+        },
+        "backhaul-cost-per-pickup": {
+          "description": "route backhaul cost per pickup",
+          "type": "string"
         }
       },
       "required": [
         "name",
         "cutoff-frequency",
-        "delivery-offset"
+        "delivery-offset",
+        "internal"
       ]
     },
     "newRoute": {


public.json
```diff
@@ -2858,12 +2858,39 @@
           "description": "number of days after cutoff before the first stop",
           "type": "integer",
           "format": "int32"
+        },
+        "internal-only": {
```
wking: Probably just `internal`, since “partially internal” isn't a thing ;).

public.json
```diff
@@ -2858,12 +2858,39 @@
...
+          "description": "boolean determining if the route is interal",
```
wking: Existing boolean docs aren't very consistent, but “{Question answered by the boolean}?  {Additional notes}” seems like a reasonable convention to pick up.  So maybe, “Is the route for internal use?  This controls whether customers can access the route via the API (they can't if internal is true).”

public.json
```diff
@@ -2858,12 +2858,39 @@
...
+          "type": "boolean"
+        },
+        "description-of-order-deadline": {
+          "description": "description of the route order deadline",
+          "type": "string"
+        },
```
wking: We don't seem to actually use this for anything:

    set([a.deadline_description for a in Route.objects.all()])
    # set([u'',
    #      u'this is only for the invoices from what the employees have already taken off the shelf',
    #      u'No description yet.',
    #      u'test',
    #      None])

The only meaningful text is the employee note for the X route, and that doesn't seem like a deadline description ;).  So I'd suggest dropping this field.

public.json
```diff
@@ -2858,12 +2858,39 @@
...
+        "backhaul-cost-per-pickup": {
+          "description": "route backhaul cost per pickup",
+          "type": "string"
+        },
+        "zip-codes-served": {
+          "description": "zip codes served for the route",
+          "type": "string"
+        },
```
wking: This sounds useful, but it looks like we don't have any useful data in that field:

    set([tuple(a.routepostalcode_set.all()) for a in Route.objects.all()])
    # set([()])

I expect we can fill it in from drop addresses along the route.  That would miss zip codes we passed through but didn't stop in, but I think we should get some workflow docs around what this is for before deciding that drop-zips aren't sufficient.

PR #40, opened 2015-10-02
public.json: Add order.fees

### LINKED ISSUE #37: API representation for order price and shipping costs
[Our current logic][1] for them incorrectly assumes the payment is order-lines + shipping, but that's not going to be true if the customer had an outstanding credit/balance that got rolled into that payment.  And checkout-payment seems to have not made it into the API spec anyway (see azurestandard/beehive@34e7a19cb for how we're setting it).  Factors to consider:

* We break shipping out on invoices, so it makes sense to have a separate `order.shipping`.  The total for an order is then (with #36) `sum(orderLines.price) + order.shipping`.
* With `order.shipping`, we have a hard time expanding to other auxiliary costs.  For example, if we charge a small-order fee, we'd need a separate field `order.small-order-fee`.  Maybe it's better to have `order.fees`, which would be an object like `{'shipping': 1.23, 'small-order': 50.0}`.  That could give UIs some forward compatibility if we add additional fee-types later (they'll probably want to map the slugs to human-readable names, but if you see a slug you haven't mapped, and you don't want to automatically map, you can always just print the slug).
* With any sort of shipping-via-an-order-property approach, it's going to be hard to update shipping after adjusting an order line.  For example, if you add a heavy order line, would you POST the line, wait for a response, and then re-GET the order just to update your shipping estimate?  We don't do shipping estimates now, but we've talked about them a lot (especially for UPS orders), so I don't want to paint ourselves into a corner here.  On the other hand, `GET /order/{id}` is pretty cheap at around 300ms (from BH-PW, but see also [these notes on Slack][2]), so maybe we don't need to embed the (expensive to generate) product at all.

[1]: https://github.com/azurestandard/azure-angular-providers/blob/v0.5.0/azure-providers.js#L719-L723
[2]: https://azurestandard.slack.com/archives/general/p1443738763003264

wking:
Builds on #39, so review/merge that first.

Existing fees in the database look like:

    {"order": 123, "type": "shipping", "amount": 0.39, "notes": "8.50%"}
    {"order": 456, "type": "small-order", "amount": 2.50, "notes": "Automatically Added"}

I'm not wild about the “Automatically Added” (do we have fees that are
manually assessed?), but having room for an optional note of some kind
makes sense.

Shipping and small-order fees are the only ones I know of, but we can
easily extend this to new fees by adding entries to orderFeeType.

The explicit orderFee.id, orderFee.order, and use of an array for
order.fees instead of an object are setting us up for an `/order-fee`
endpoint if we decide to move fee management into the API
(e.g. letting internal UIs POST fees to orders, or DELETE fees from
orders).

I haven't required orderFee.id at the moment, because we currently
handle [UPS shipping outside of our fee model][1].  If we do decide to
add an API for manipulating order-fee instances we'll need to collapse
that into our existing order-fee table.

Survey of Beehive's fee structure:

* Invoice template uses [UPS-specific shipping][2] and then [general
  fees][3].
* [Fee types][4] (Beehive uses “reason”, but I've gone with “type” in
  the API to match the existing paymentMethod.type and telephone.type)
  are:
  * `ORDER_FEE_REASON_LATE` (never used.  `git grep
    ORDER_FEE_REASON_LATE` turns up no consumers, and
    `OrderFee.objects.filter(reason=ORDER_FEE_REASON_LATE).count()`
    returns zero.
  * `ORDER_FEE_REASON_SHIPPING_TRUCK` is a [per-route shipping fee][5]
    (so “shipping”).
  * `ORDER_FEE_REASON_SMALL_ORDER` is a [charge for small orders we
    ship ourselves][6] (so “small-order”).
  * `ORDER_FEE_REASON_SHIPPING_UPS` is a [charge for small UPS
    orders][7] (so “small-order”).

Fixes #37.

[1]: https://github.com/azurestandard/beehive/blob/554325e15c86087d4dc79f6ffab1687a72a7c2e2/apps/orders/models.py#L1766
[2]: https://github.com/azurestandard/beehive/blob/554325e15c86087d4dc79f6ffab1687a72a7c2e2/apps/customers/templates/customers/website/invoice.html#L13
[3]: https://github.com/azurestandard/beehive/blob/554325e15c86087d4dc79f6ffab1687a72a7c2e2/apps/customers/templates/customers/website/invoice.html#L14
[4]: https://github.com/azurestandard/beehive/blob/554325e15c86087d4dc79f6ffab1687a72a7c2e2/apps/orders/constants.py#L37-L46
[5]: https://github.com/azurestandard/beehive/blob/554325e15c86087d4dc79f6ffab1687a72a7c2e2/apps/orders/models.py#L1018-L1023
[6]: https://github.com/azurestandard/beehive/blob/554325e15c86087d4dc79f6ffab1687a72a7c2e2/apps/orders/models.py#L1199-L1203
[7]: https://github.com/azurestandard/beehive/blob/554325e15c86087d4dc79f6ffab1687a72a7c2e2/apps/orders/models.py#L1195-L1198

### DIFF
--- a/public.json
+++ b/public.json
@@ -3419,6 +3419,13 @@
           "description": "before shipping, the customer can use this to associate a payment-method with the order.  After shipping, it contains information about the charged payment",
           "$ref": "#/definitions/payment"
         },
+        "fees": {
+          "description": "additional charges beyond the order-line items themselves (estimated until picking time)",
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/orderFee"
+          }
+        },
         "drop": {
           "description": "drop the order is destined for (unset for UPS orders)",
           "type": "integer",
@@ -3490,6 +3497,47 @@
         "quantity-ordered",
         "weight",
         "volume"
+      ]
+    },
+    "orderFee": {
+      "description": "An additional charge associated with an order.",
+      "type": "object",
+      "properties": {
+        "id": {
+          "type": "integer",
+          "format": "int64"
+        },
+        "order": {
+          "description": "Order that the fee belongs to",
+          "type": "integer",
+          "format": "int64"
+        },
+        "type": {
+          "description": "An array of properties for this number",
+          "$ref": "#/definitions/orderFeeType"
+        },
+        "amount": {
+          "description": "Dollar value for the fee",
+          "type": "number",
+          "format": "float"
+        },
+        "notes": {
+          "description": "Additional information about the reason or amount for the fee",
+          "type": "string"
+        }
+      },
+      "required": [
+        "order",
+        "type",
+        "amount"
+      ]
+    },
+    "orderFeeType": {
+      "description": "A property for an order fee",
+      "type": "string",
+      "enum": [
+        "shipping",
+        "small-order"
       ]
     },
     "accountEntry": {
--- a/public.json
+++ b/public.json
@@ -3415,6 +3415,10 @@
           "type": "string",
           "format": "date-time"
         },
+        "checkout-payment": {
+          "description": "before shipping, the customer can use this to associate a payment-method with the order.  After shipping, it contains information about the charged payment",
+          "$ref": "#/definitions/payment"
+        },
         "drop": {
           "description": "drop the order is destined for (unset for UPS orders)",
           "type": "integer",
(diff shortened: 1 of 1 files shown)


tompetersjr:
This looks like a reasonable approach to me.

PR #41, opened 2015-10-08
Update route endpoints to handle recently added data

colin-merc-tech:


wking:
I tweaked the commit message to clarify what was happening (we already had an updateRoute endpoint, but the Swagger spec for it cheated, with the cheat explained in a description).

PR #43, opened 2015-10-22
Add `updateDrop` definition for PUT requests.

caseybessette:


public.json
```diff
@@ -2969,6 +2969,74 @@
...
+        },
+        "geo": {
+          "description": "recalculated from address if not included",
+          "$ref": "#/definitions/geo"
+        },
+        "active": {
+          "description": "whether or not this drop is actively ordering",
+          "type": "boolean"
+        },
+        "exclusivity": {
```
wking: We don't actually care about this (it's [calculated][1] [from][2] [other fields][3]).  I'd like to make it a database field (azurestandard/beehive#987, azurestandard/beehive#1048), but until we do that we should probably leave it out of the API.

[1]: https://github.com/azurestandard/beehive/blob/8913c96bd061c22ea28f0ec5e7697601466618d8/apps/api/views/drops.py#L166
[2]: https://github.com/azurestandard/beehive/blob/8913c96bd061c22ea28f0ec5e7697601466618d8/apps/api/views/drops.py#L98-L100
[3]: https://github.com/azurestandard/beehive/blob/8913c96bd061c22ea28f0ec5e7697601466618d8/apps/drops/templatetags/pretty_lights.py#L11-L46

public.json
```diff
@@ -2969,6 +2969,74 @@
...
+        "fees": {
+          "$fees": "#definitions/dropFees"
+        },
+        "notes": {
+          "description": "free-form Markdown notes for any drop information that doesn't fit into an existing field",
+          "type": "string"
+        }
+      },
+      "required": [
+        "id",
```
wking: Sometimes we recycle `update*` for the creation POST, in which case we won't have an ID.  Even if we don't we'll ignore the `update*` ID field in most cases (but maybe not for routes, see azurestandard/beehive@5f80fe9), because we don't allow users to change the ID.  So I wouldn't require it here (for PUTs, it will be in the URL anyway).

PR #44, opened 2015-10-26
Update trips spec to include new endpoints

colin-merc-tech:


public.json
```diff
@@ -628,6 +660,75 @@
...
+        "summary": "Updates a trip for a route, if the user has permissions",
+        "operationId": "updateTrip",
+        "tags": [
+          "trip"
+        ],
+        "parameters": [
+          {
+            "name": "id",
+            "in": "path",
+            "description": "ID of trip to fetch",
```
wking: s/fetch/update/

public.json
```diff
@@ -628,6 +660,75 @@
...
+            "required": true,
+            "type": "integer",
+            "format": "int64"
+          },
+          {
+            "name": "trip",
+            "in": "body",
+            "description": "Trip to create",
```
wking: s/create/update/

public.json
```diff
@@ -628,6 +660,75 @@
...
+          },
+          "default": {
+            "description": "unexpected error",
+            "schema": {
+              "$ref": "#/definitions/errorModel"
+            }
+          }
+        }
+      },
+
```
wking: No need for a blank line, and we don't currently use them to space out other endpoint docs.

public.json
```diff
@@ -628,6 +660,75 @@
...
+        "summary": "Deletes a trip for a route, if the user has permissions",
+        "operationId": "deleteTrip",
+        "tags": [
+          "trip"
+        ],
+        "parameters": [
+          {
+            "name": "id",
+            "in": "path",
+            "description": "ID of trip to fetch",
```
wking: s/fetch/delete/

public.json
```diff
@@ -3070,6 +3171,55 @@
...
+        "backhaul-end": {
+          "description": "estimated date for the last backhaul",
+          "type": "string",
+          "format": "date-time"
+        },
+        "warehouse-arrival": {
+          "description": "estimated date for warehouse arrival",
+          "type": "string",
+          "format": "date-time"
+        }
```
wking: Some thoughts on these fields:

* Any fields we have in `updateTrip` we probably want in the returned `trip`, otherwise editing will be difficult.
* Now that we're shifting to a unified drop/pickup approach, do we still need separate `delivery-*` and `backhaul-*` variables?  Grepping through Beehive, it seems like we only use `warehouse_arrive` as one of the cases in `Purchase.expected_arrival` (so only post-pickup).  If trucks are going to be picking up POs on both inbound and outbound legs (or outbound and inbound legs), it seems like we might need two of these dates.  I think it would be easier to just treat the warehouse as another stop in the route/trip (and that scales better if we add additional warehouses in the future).
* I'd rather calculate the `*-end` time directly from `*-start` + max offset.  So set it in `trip`, for convenience, but don't pass it with `updateTrip`.

wking: Actuallly, we do need `delivery-start`, since [that's the anchor for our route-stop offsets][1].  Other than that, I still agree with my earlier assessment.

[1]: https://github.com/azurestandard/api-spec/blob/a5389dd7d770f4e511d36ee8c33f04f27f140661/public.json#L3182-L3186

wking: A lot of this was discussed on Slack today (starting [here][1]).  I think we've resolved everything in favor of just storing `delivery-start` in the API, but we still haven't figured out how to handle `warehouse_arrive` (and are trying to close the loop with @JBaisinger on that).

[1]: https://azurestandard.slack.com/archives/general/p1447962718001316

tompetersjr: JBaisinger got back to us via email and we are going to allow for the manual entry of Delivery End and Backhaul End.  James said that they could be computed most of the time but there will be cases where we want to manually enter this in.  So we could auto fill Delivery End with the lost stop on the trop and Backhaul End with the last pickup on the trip.  However the UI still needs to allow the user to override the value.

This should free up all outstanding issues around this so that we can merge it to unblock heuscher.

wking: On Tue, Nov 24, 2015 at 04:58:04PM -0800, Tom Peters wrote:
> JBaisinger got back to us via email and we are going to allow for
> the manual entry of Delivery End and Backhaul End.

Yeah, it sounds like the current requirements are “expose all these
fields”.  So the only remaining issue from [1] is:

* Any fields we have in updateTrip we probably want in the returned
  trip, otherwise editing will be difficult.

[1]: https://github.com/azurestandard/api-spec/pull/44/files#r43014832

tompetersjr:
@colin-merc-tech:  Want to address wking's last comment and we can get this rolled out.  This one is blocking a few things so I would like to get this one resolved and deployed.  Thanks.

colin-merc-tech:
I believe that last issue was handled here: https://github.com/azurestandard/api-spec/commit/d155d881a88c3315d4b0c6deaa414ffe3943bcc5 @wking let me know if that's not the case.

PR #45, opened 2015-10-26
Add pickup ID to stop return schema

colin-merc-tech:


wking:
Cherry-picked onto master and merged with the following two minor
tweaks:

* Shifted it in the JSON so it sits next to the related .drop
  attribute.
* Removed 'drop' from the required stop attributes, to allow
  pickup-only stops.

PR #49, opened 2015-10-31
Logistics APIs for purchases and pickups

wking:
This reverts #45 and takes another pass at the purchase/pickup
modeling now that I understand it better.  See [here][1] for an
overview of our current database representation.  The API I'm
proposing makes the following adjustments to those database models:

* Squashes `BackhaulPickup` and `Purchase` into the purchase API (like
  `Order` and `OrderShipment` are squashed into `Order`).
  `BackhaulPickup` is one-to-one with `Purchase`, so this squash can
  be lossless.

* Exposes `VendorContact` through the pickup API.  I expect there will
  be the same sort of confusion here that we got from the `Product` →
  packaged-product and `PieceMeta` → product renaming, but I think
  ‘pickup’ is a clear echo of ‘drop’ and ‘vendor-contact’ doesn't
  strike me as a physical location that could be shared by several
  vendors.  This also shifts the route association

Further details in the commit messages.

I'd *like* to have both route-drop and route-pickup associations
managed via the route-stop API (which is why it's not the route-drop
API ;), but that may be too much of a stretch for our current database
models.  On the other hand (as I point out [here][1], the current
database models for pickup logistics don't make all that much sense,
and we have more flexibility changing logistics modeling and UIs than
we do for changing customer-facing modeling and UIs.

[1]: https://github.com/azurestandard/beehive/blob/tk/model-docs/docs/modeling/logistics.md

### DIFF
--- a/public.json
+++ b/public.json
@@ -2646,6 +2646,123 @@
         }
       }
     },
+    "/vendors": {
+      "get": {
+        "summary": "Returns all vendors from the system that the user has access to",
+        "description": "Vendors are ordered for increasing ID (creation time)",
+        "operationId": "findVendors",
+        "tags": [
+          "vendor"
+        ],
+        "parameters": [
+          {
+            "name": "active",
+            "in": "query",
+            "description": "only return (in)active vendors",
+            "required": false,
+            "type": "boolean"
+          },
+          {
+            "name": "manager",
+            "in": "query",
+            "description": "person IDs to filter by, selecting vendors that these vendor employees manage",
+            "required": false,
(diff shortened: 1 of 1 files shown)


wking:
Actually, I just realized that the `VendorContact` → pickup quasi-rename means #45 is ok, so I've rerolled this branch to not revert it.  After this PR, pickup will just be a drop-like address that's reused between trips.

wking:
Rebased around #40 (trivial conflicts due to `vendor` and `orderFee` being added in the same part of the file).

PR #51, opened 2015-11-13
Add 'POST /order/{id}/shipment': Allow customers to see order-lines that will be removed when switching to a parcel carrier cart.

wking:
So folks who want to see what order-lines get removed can do that
without polling and comparing.  More background in
azurestandard/beehive#1202, especially
azurestandard/beehive@f92ed0573c.

### DIFF
--- a/public.json
+++ b/public.json
@@ -1641,7 +1641,7 @@
     },
     "/order/{id}": {
       "get": {
-        "summary": "Returns a order based on a single ID",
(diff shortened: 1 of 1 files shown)


caseybessette:
Slack discussion on this starting here https://azurestandard.slack.com/archives/general/p1457984165000052

wking:
Obsoleted by #126, which covers side-effects due to any order change (e.g. placing an order).

PR #52, opened 2015-11-14
public.json: Add 'POST /password/reset' and 'POST /password/confirm'

wking:
Reading through azurestandard/beehive#1330 reminded me that we didn't
have a way to reset passwords with the new API-based website, so I'm
adding one here.

POSTing a known email address to /password/reset will cause a message
to be sent to that address.  The email will contain a link with a
reset token, and that token can be POSTed to /password/confirm to set
a new password.  This base-url / token approach is just like the one
used for registration, and we support both POST and GET confirmations
for the same reasons (see 182d6c9, public.json: Add 'GET
/registration/confirm?token=...', 2015-09-03, #19).

/password/confirm returns the person whose password was reset, so the
UI can display their name, etc. in a successful-reset page.

### DIFF
--- a/public.json
+++ b/public.json
@@ -3029,6 +3029,109 @@
         }
       }
     },
+    "/password/reset": {
+      "post": {
+        "summary": "Initiate a password reset, to be completed by a call to /password/confirm",
+        "operationId": "resetPasswordRequest",
+        "tags": [
+          "password"
+        ],
+        "parameters": [
+          {
+            "name": "reset",
+            "in": "body",
+            "required": true,
+            "type": "object",
+            "schema": {
+              "$ref": "#/definitions/passwordResetRequest"
+            }
+          }
+        ],
+        "responses": {
+          "204": {
+            "description": "Password-reset email sent (if the registration email corresponded to an existing person.  Otherwise, you'll still get a 204 to avoid leaking information about member email addresses)"
+          },
+          "default": {
+            "description": "unexpected error",
+            "schema": {
+              "$ref": "#/definitions/errorModel"
+            }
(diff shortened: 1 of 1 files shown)


public.json
```diff
@@ -2593,6 +2593,108 @@
...
+            }
+          },
+          "default": {
+            "description": "unexpected error",
+            "schema": {
+              "$ref": "#/definitions/errorModel"
+            }
+          }
+        }
+      }
```
caseybessette: Syntax error here.

public.json
```diff
@@ -3832,6 +3934,44 @@
...
+          "type": "string",
+          "format": "email"
+        }
+      },
+      "required": [
+        "base-url",
+        "email"
+      ]
+    },
+    "passwordReset": {
```
caseybessette: Maybe call this `passwordResetConfirmation`?

caseybessette:
@wking Trevor, this looks good to me.  I can make the fixes I noted/suggested, fix the merge conflict, and merge it.  In fact, I will do that sometime tomorrow barring any objections.

wking:
On Wed, Feb 24, 2016 at 02:09:57PM -0800, Casey M. Bessette wrote:
> I can make the fixes I noted/suggested, fix the merge conflict…

Suggestions sound good, so I just did all that ;).

wking:
On Wed, Feb 24, 2016 at 03:02:11PM -0800, W. Trevor King wrote:
> Suggestions sound good, so I just did all that ;).

And I just pushed 95bc594 → a94f2e2 so we weren't using the same
description for both new models.

PR #53, opened 2015-12-14
add fields to packaged-product that are required for the categories app.

MercenaryJosh:
adds the changes to the packaged-products api response introduced in #29673dc.

### DIFF
--- a/public.json
+++ b/public.json
@@ -3458,6 +3458,12 @@
         "bargain-bin-notes": {
           "description": "Bargain-bin notes for this packaged product",
           "type": "string"
+        },
+        "product": {
+          "description": "the name and brand of the product that this packaged-product is associated with",
+          "schema": {
+            "$ref": "#/definitions/packagedProductProduct"
+          }
         }
       },
       "required": [
@@ -3466,6 +3472,22 @@
         "price",
         "tags"
       ]
+    },
+    "packagedProductProduct": {
+      "description": "the name and brand of a product that a packaged-product is associated with",
+      "type": "object",
+      "properties": {
+        "name": {
+          "description": "the product name",
+          "type": "string",
+        },
+        "brand": {
+          "description": "the brand that this product is associated with",
+          "schema": {
+            "$ref": "#/definitions/brand"
+          }
+        }
+      }
     },
     "packagedProductTag": {
       "description": "a type of packaged product or other boolean marker",


MercenaryJosh:
> That cuts both ways, since names are attached to products and not
packaged-products.  Maybe it would help the intuitiveness if we just
drop /packaged-products and /packaged-product/{code} from the API?  I
can't think of a situation off the top of my head where they'd be a
better choice than their product analogs.

The data returned from the /product endpoints includes a "packaging" field that is an array of packaged-products. I do not see the conceptual difference between that and the data from the /packaged-product endpoints having a "product" field that includes information about the product that it is associated with.

Removing the /packaged-product endpoint entirely does not seem feasible. There needs to be a way to add/remove packaged-products to/from categories, set the packaged-product's primary category, and I expect that there will need to be a way to add/edit packaged-products.

jeffstrickler:
So, the above conversation seems to revolve around two points:

1) Josh has a requirement from Susan to set a category for each type of packaging for a given product. 

2) By the above conversation, Trevor appears to have a preference that product is a public API and packaged-products is perhaps intended to be a private API?

So here are the parts I am trying to understand:

a) Since the requirement is to set categories at a packaging level, this seems to be a natural fit with packaged-products

b) I am unclear why there would be a preference for clients to use the product endpoint over the packaged-products endpoint, when they are both currently exposed publicly. 

c) The categories app has been built around packaged-products because of this requirement, and while it might not be a huge change to use the products endpoint instead, it isn't immediately clear to be what the net-gain might be, especially if time is of the essence. 

d) Since the data needs to update immediately, it isn't coming from algolia. So for example, there isn't a gain in using one endpoint over the other (IE: it isn't like we can get rid of one of the indexes by switching use over to another).

tompetersjr:
> a) Since the requirement is to set categories at a packaging level, this seems to be a natural fit with packaged-products

That sounds reasonable to me.

> b) I am unclear why there would be a preference for clients to use the product endpoint over the packaged-products endpoint, when they are both currently exposed publicly.

Sounds like there is some redundancy here and Trevor agrees based on 

> I can't think of a situation off the top of my head where they'd be a better choice than their product analogs.

However Jeff also makes a good point here.

> c) The categories app has been built around packaged-products because of this requirement, and while it might not be a huge change to use the products endpoint instead, it isn't immediately clear to be what the net-gain might be, especially if time is of the essence.

Time is of the essence.  The endpoint already exists for some reason.  I think we should use it.

> d) Since the data needs to update immediately, it isn't coming from algolia. So for example, there isn't a gain in using one endpoint over the other (IE: it isn't like we can get rid of one of the indexes by switching use over to another).

I'd like to go the route of using packaged-products unless there is significant reasoning to take the time now to rip out packaged-products and convert anything that is using it over to the products api.  Right now I see the reasoning to make the API simpler but I think the timeline wins over that at this point in time.

wking:
On Tue, Dec 15, 2015 at 09:27:29PM -0800, MercenaryJosh wrote:
> The data returned from the /product endpoints includes a "packaging"
> field that is an array of packaged-products. I do not see the
> conceptual difference between that and the data from the
> /packaged-product endpoints having a "product" field that includes
> information about the product that it is associated with.

Fair enough.  I think my concern is mostly about creating loops of
such additional inclusions, although there's no technical reason for
the loops to be a problem (although the packaging lookups currently
about double the latency of /packages requests, so there's a
performance cost for any additional inclusions that can't be handled
with select_related).

> Removing the /packaged-product endpoint entirely does not seem
> feasible. There needs to be a way to add/remove packaged-products
> to/from categories, set the packaged-product's primary category, and
> I expect that there will need to be a way to add/edit
> packaged-products.

Yeah, /packaged-product/{packaged_product_code}/category/{category_id}
needs to stay.  I was just talking about deprecating the
/packaged-products and /packaged-product/{code} endpoints.  I'm not
sure what the primary-category API looks like.  Is that in a PR/branch
somewhere?

Anyhow, Tom's point about just rolling with the augmented
packagedProduct model until we iron the bugs out of /products makes
sense.  I'd recommend:

* dropping e96bf5df7 (I think we agree on that [1]), and
* squashing 5e84eedf11 into 2ee7507efc.

[1]: https://github.com/azurestandard/api-spec/commit/e96bf5df7ae46e92483ecbf1413045e7a7455ce3#commitcomment-14968880

MercenaryJosh:
> I think my concern is mostly about creating loops of
such additional inclusions

The implementation of this in azurestandard/beehive@29673dc only includes the product field from the /packaged-products apis, not the /products apis, so looping is not a problem.

>Anyhow, Tom's point about just rolling with the augmented
packagedProduct model until we iron the bugs out of /products makes
sense.  I'd recommend:
* dropping e96bf5d (I think we agree on that [1]), and
* squashing 5e84eed into 2ee7507.

Sounds good. If you want to do that and then merge the pull request that would be great. Thanks.

PR #54, opened 2015-12-17
Primary Category

MercenaryJosh:
Adds fields to the category model and a query parameter to /packaged-products in order to add the new features and requirements to the Category model and in the categories app.

### DIFF
--- a/public.json
+++ b/public.json
@@ -3464,6 +3464,11 @@
           "schema": {
             "$ref": "#/definitions/packagedProductProduct"
           }
+        },
+        "primary-category": {
+          "description": "the id of the primary category for this packaged-product",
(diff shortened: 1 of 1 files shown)


wking:
On Wed, Dec 16, 2015 at 04:22:20PM -0800, MercenaryJosh wrote:
>   * add "missing-primary-category" query parameter to /packaged-products
>   * add "is_primary" field to the category model
>   * add "active" field to the category model
>   * add "keywords" field to the category model
>   * add "primary-category" field to packaged-product model

This is beautiful, thanks :).  I've left some comments on the
individual additions where I think there's a cleaner approach or the
motivation wasn't clear to me.

PR #55, opened 2016-01-08
add endpoints for frequently asked questions

MercenaryJosh:


public.json
```diff
@@ -2983,6 +2983,203 @@
...
+        "default": {
+          "description": "unexpected error",
+          "schema": {
+            "$ref": "#/definitions/errorModel"
+          }
+        }
+      },
+      "put": {
+        "summary": "change the order in which the faqs are displayed",
+        "operationId": "reorderFaqs",
```
wking: The ordering is the only thing we can allow people to adjust now, but I'd avoid baking it into the UI so strongly (e.g. I expect updating questions and answers will be requested features soonish, because nothing is ever perfect out of the gate).  So can we stick to the usual `updateFaq` operation ID with an `updateFaq` model definition?  Then we can 403 (for now) anyone who tries to change the question or answer.

public.json
```diff
@@ -4409,6 +4606,35 @@
...
+      "properties": {
+        "id": {
+          "type": "integer",
+          "format": "int64"
+        },
+        "question": {
+          "type": "string"
+        },
+        "answer": {
+          "type": "string"
```
wking: A description outlining the format and size here would be nice.  It it HTML, Markdown, plain text?  A few sentences?  Paragraphs?  As long as you want?

public.json
tompetersjr: We will want ordering.  And we will be able to update FAQ's via the UI here: https://internal-website.azurestandard.com/#/cms/.

public.json
```diff
@@ -4409,6 +4606,35 @@
...
+    "faq": {
+      "description": "frequently asked question",
+      "type": "object",
+      "properties": {
+        "id": {
+          "type": "integer",
+          "format": "int64"
+        },
+        "question": {
+          "type": "string"
```
wking: I'd guess this is a short sentence in plain text (no markup) ending in a question mark.  If that guess isn't right, a description outlining the intended format, size, and terminal-character flexibility would be nice.

public.json
```diff
@@ -2983,6 +2983,203 @@
...
+            "in": "query",
+            "description": "offset to the first result to return.  Use negative numbers to offset from the end of the result list.",
+            "required": false,
+            "type": "integer",
+            "format": "int32"
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "array of faq response",
```
wking: nit: I prefer “FAQ” to “faq” for the acronym, but I'm sure folks will get the idea either way ;).

public.json
wking: Err, I see you *have* an updateFaq operation in the singular endpoint.  So maybe just remove this endpoint entirely?

public.json
```diff
@@ -4409,6 +4606,35 @@
...
+        },
+        "answer": {
+          "type": "string"
+        },
+        "order": {
+          "type": "integer",
+          "format": "int64"
+        }
+      }
+    },
```
wking: You probably want a `required` section pointing out that all of these fields will always be set in responses we send out.  And then a separate `updateFaq` model where everything is required except for `id`, which we can use for POST/PUT endpoints.

public.json
wking: On Fri, Jan 08, 2016 at 08:23:06AM -0800, Tom Peters wrote:
> > +      "put": {
> > +        "summary": "change the order in which the faqs are displayed",
> > +        "operationId": "reorderFaqs",
> 
> We will want ordering…

This is PUT /faqs with '{1: 2, 2: 1}' (id: order).  Why not just have
a few PUT /faq calls that set the order number per-entry?  That's how
we handle address ordering now (with address.preference), and I don't
expect FAQ shuffling to happen so frequently that we need the
implemenation/maintenance cost of a custom endpoint.

MercenaryJosh: 1. This is not a custom endpoint, it is handling the PUT method on the /faqs endpoint.
2. It is much simpler and more efficient for the browser to send one request when the questions are reordered than to send over a dozen.

wking: On Fri, Jan 08, 2016 at 01:04:57PM -0800, MercenaryJosh wrote:
> > +        "operationId": "reorderFaqs",
> 
> 1. This is not a custom endpoint, it is handling the PUT method on
>    the /faqs endpoint.

The semantics for PUT are [1]:

  The PUT method requests that the state of the target resource be
  created or replaced with the state defined by the representation
  enclosed in the request message payload.  A successful PUT of a
  given representation would suggest that a subsequent GET on that
  same target resource will result in an equivalent representation
  being sent in a 200 (OK) response.

We're probably not completely RFC-compliant, but a “reorder the
entries” endpoint is a pattern that we don't have in an existing
endpoint.

> 2. It is much simpler and more efficient for the browser to send one
> request when the questions are reordered than to send over a dozen.

For simplicity, the JavaScript is something like:

  locallyOrderedFaqs.forEach(function (faq, index) {
    faq.order = index;
    faq.$save().catch(alertService.catchNetwork('faq'));
  });

which doesn't seem too bad to me.

For efficiency, I think FAQ-reordering will be much less frequent than
address-reordering.  Can we leave this out for now, and restore it
if/when we see performance issues with the “PUT each FAQs with its new
order” approach?

[1]: http://tools.ietf.org/html/rfc7231#section-4.3.4

wking:
With 23260908, I think the only thing this is missing is [`required` sections][1].

[1]: https://github.com/azurestandard/api-spec/pull/55#discussion_r49207247

MercenaryJosh:
Added `required` section. Can this be merged?

wking:
On Tue, Jan 19, 2016 at 02:33:50AM -0800, MercenaryJosh wrote:
> Added `required` section. Can this be merged?

Rebased onto master and merged.

PR #57, opened 2016-01-08
Add product's nutrition-facts.

caseybessette:


public.json
```diff
@@ -3604,6 +3607,45 @@
...
       ]
     },
+    "nutritionFacts": {
+      "description": "nutrition facts",
+      "type": "object",
+      "properties": {
+        "headers": {
+          "type": "array",
+          "items": {
+            "description": "nutrition facts headers (e.g., serving size, calories)",
```
wking: nit: a description about the *array* (plural headers) would go in the headers property, not in headers.items (e.g. see order.fees).  A description about an *entry* in that array (singular header) would go here (e.g. see category.keywords).  In this case, I'd recommend:

    "items": {
        "description": "a nutrition-fact header (e.g. \"Serving Size 1 cake (19g)\" or \"Calories 60\", ...)",
        …

caseybessette: Good point. Updating this now.

public.json
```diff
@@ -3604,6 +3607,45 @@
...
+            }
+        }
+      }
+    },
+    "nutritionFactsDailyValue": {
+      "description": "nutrition facts daily values",
+      "type": "object",
+      "properties": {
+        "indent": {
+          "description": "if true, this label should be indented",
```
wking: To clarify, I'd add:

> (e.g. an indented \"Dietary Fiber\" entry might follow a non-indented \"Total Carbohydrate\" entry)

public.json
```diff
@@ -3604,6 +3607,45 @@
...
+          "type": "boolean"
+        },
+        "label": {
+          "description": "daily value label (e.g., cholesterol, sodium)",
```
wking: Pulling examples from PieceMeta 10425, we would have:

    "description": "daily value label (e.g. \"Total Carbohydrate 14g\", \"Calcium\")",

public.json
```diff
@@ -3604,6 +3607,45 @@
...
+          "type": "string"
+        },
+        "percentage": {
+          "description": "daily value percentage (e.g., 10.0)",
+          "type": "float"
```
wking: I think we want strings here.  For example, in my local DB I have PieceMeta 9392 (inactive, but still) with:

    {'indent': False, 'label': u'Total Carbohydrate 1 g', 'percentage': u'<1'}

caseybessette:
Updated per your comments.  Not sure what possessed me to do the nested items.items. :)

PR #59, opened 2016-01-13
Add verified to stop definition

colin-merc-tech:
ref https://github.com/azurestandard/beehive/pull/1449

### DIFF
--- a/public.json
+++ b/public.json
@@ -3429,6 +3429,10 @@
           "description": "actual stop time (set after the stop is reached)",
           "type": "string",
           "format": "date-time"
+        },
+        "verified": {
+          "description": "whether the stop is verified or not",
+          "type": "boolean"
         }
       },
       "required": [


wking:
Can we sketch out some workflows around this setting here.  When will
users want to access unverified stops?  Why don't we just delete those
stops at verification time?

Also, we should update the stop descriptions to explain that they're
being created at trip-creation time [1].

[1]: https://azurestandard.slack.com/archives/general/p1450221251003741

colin-merc-tech:
I'm not certain what you want for workflows but having stops created before cutoff was a request from the logistics team, though Jeff may be better suited to explain why they want that. As for deleting unverified stops, keeping them just feels like better practice. It allows us to maintain historical record regarding drops that existed on the route at the time of the trip and provides a visual cue that can be used to show drops that don't have orders. 

With that said, if it's something you want to pursue further, perhaps we can talk to logistics and see if they have a preference.

wking:
On Fri, Jan 22, 2016 at 08:17:23AM -0800, colin-merc-tech wrote:
> I'm not certain what you want for workflows but having stops created
> before cutoff was a request from the logistics team, though Jeff may
> be better suited to explain why they want that.

Creating stops at trip-creation time makes sense, because you need
that for adjusting per-stop times before verification [1].  My concern
is just about whether we need to keep unverified stops after
verification time.

> As for deleting unverified stops, keeping them just feels like
> better practice. It allows us to maintain historical record
> regarding drops that existed on the route at the time of the trip
> and provides a visual cue that can be used to show drops that don't
> have orders.

We already have that historical record in DropPointTrip.  Until we
have a clear use-case for preserving that record, I'd rather not add
it to the API ;).

> With that said, if it's something you want to pursue further,
> perhaps we can talk to logistics and see if they have a preference.

That sounds like a good way to figure out any use cases for “this stop
was considered for trip 123 but ended up not being chosen”.  I'm also
fine going with “delete unverified stops at verification time” now and
revisiting if/when the archival use case comes up.

[1]: https://azurestandard.slack.com/archives/general/p1450220697003732

colin-merc-tech:
I'll remove the field and switch it over to deleting them at verification. Once that's removed I'll close this PR, as there will be no actual changes to the API.

wking:
On Fri, Jan 22, 2016 at 12:14:24PM -0800, colin-merc-tech wrote:
> I'll remove the field and switch it over to deleting them at
> verification. Once that's removed I'll close this PR, as there will
> be no actual changes to the API.

We still need to update some descriptions.  I'll file a PR and assign
it to you.

colin-merc-tech:
Sounds good. Closing this PR.

PR #61, opened 2016-01-20
Add remaining drop settings & preferences.

caseybessette:
Spec changes for https://github.com/azurestandard/beehive/pull/1482

### DIFF
--- a/public.json
+++ b/public.json
@@ -3286,6 +3286,78 @@
           "description": "whether or not this drop is actively ordering",
           "type": "boolean"
         },
+        "accounts-payable-contact": {
+          "description": "name of the drop employee Azure's accounts-payable should contact, if different from the primary contact",
+          "type": "string"
+        },
+        "after-hours-phone": {
+          "description": "E.123 telephone number Azure should call if we need to contact the drop after business hours",
+          "type": "string"
+        },
+        "business-hours": {
+          "description": "regular business hours for the drop (e.g. \"8:30-6 Mon-Fri, 9-5 Sat, closed Sun\")",
+          "type": "string"
+        },
+        "business-type": {
+          "description": "what type of business do you run?",
+          "type": "array",
+          "items": {
+            "type": "string",
+            "enum": [
+              "clinics-practitioner",
+              "manufacturing-business",
+              "online-store",
+              "restaurant",
+              "retail-store",
+              "seasonal",
+              "wholesale-distributor"
+            ]
+          }
+        },
+        "buyer": {
+          "description": "name of the drop employee who usually places orders for this drop, if different from the primary contact",
+          "type": "string"
+        },
+        "pickup-contact": {
+          "description": "do customers need to contact you to pick up their orders?  Retail preference.  Only available when how-members-receive is \"can-hold-sometimes\" or \"can-hold\"",
+          "type": "string",
+          "enum": [
+            "yes",
+            "no",
+            "first-time-only"
+          ]
+        },
+        "contact-person": {
+          "type": "string"
+        },
+        "contact-type": {
+          "description": "how can customers contact you?",
+          "type": "array",
+          "items": {
+            "type": "string",
+            "enum": [
+              "phone",
+              "email",
+              "other"
+            ]
+          }
+        },
+        "contact-type-other-notes": {
+          "description": "description of the alternative contact method.  Available when contact-type includes \"other\"",
+          "type": "string"
+        },
+        "contact-type-phone": {
+          "description": "E.123 telephone number customers should call to contact this drop.  Available when contact-type includes \"phone\"",
+          "type": "string"
+        },
+        "directions": {
+          "description": "driving directions for a truck approaching this drop (e.g. avoiding low bridges, tight turns, etc. on more obvious routes)",
+          "type": "string"
+        },
+        "driver-callable-phone": {
+          "description": "E.123 telephone number for a driver approaching this drop",
+          "type": "string"
+        },
         "exclusivity": {
           "type": "string",
           "enum": [
@@ -3294,6 +3366,40 @@
             "closed"
           ]
         },
+        "food-stamps": {
+          "description": "does your drop accept food stamps?  Business preference.",
+          "type": "string",
+          "enum": [
+            "yes",
+            "no",
+            "limited"
+          ]
+        },
+        "food-stamps-limited-notes": {
+          "description": "description of limited food-stamp acceptance.  Business preference.  Only available when food-stamps is \"limited\"",
+          "type": "string"
+        },
+        "hold-time": {
+          "description": "how long will you hold a customer's order?  For example, \"24 hours\", \"three days for dry, frozen by arrangement\", etc.  Retail preference.  Only available when how-members-receive is \"can-hold-sometimes\" or \"can-hold\"",
+          "type": "string"
+        },
+        "how-members-receive": {
+          "description": "how do your members receive their items?  Retail preference",
+          "type": "string",
+          "enum": [
+            "must-show",
+            "can-hold-sometimes",
+            "can-hold"
+          ]
+        },
+        "invoice-always": {
+          "description": "always print paper invoices, regardless of the customer's invoice-delivery preference",
+          "type": "boolean"
+        },
+        "okay-to-text-phone": {
+          "description": "can the driver send text messages to the driver-callable phone?",
+          "type": "boolean"
+        },
         "parking": {
           "description": "parking location ownership for this drop",
           "type": "string",
@@ -3301,8 +3407,13 @@
             "home",
             "business",
             "non-profit",
-            "public"
+            "public",
+            "other-building"
           ]
+        },
+        "special-orders": {
+          "description": "do you take special orders for Azure products from customers?  Wholesale preference",
+          "type": "boolean"
         },
         "storage": {
           "description": "storage facilities available at this drop",
@@ -3349,6 +3460,14 @@
         "notes": {
           "description": "free-form Markdown notes for any drop information that doesn't fit into an existing field",
           "type": "string"
+        },
+        "sells-finished-goods-to-azure": {
+          "description": "do you also sell your finished good to Azure?  Only available if business-type includes \"manufacturing-business\"",
+          "type": "boolean"
+        },
+        "uses-azure-to-drop-ship": {
+          "description": "Do you drop ship Azure orders (have Azure ship directly to your customers)?  Only available if business-type includes \"online-store\"",
+          "type": "boolean"
         }
       },
       "required": [


public.json
```diff
@@ -3183,6 +3183,79 @@
           "description": "whether or not this drop is actively ordering",
           "type": "boolean"
         },
+        "accounts-payable-contact": {
+          "type": "string"
```
wking: Descriptions for these fields would be nice.  We don't have much detail in DropPointBizManagerPreferences, etc., but I'll work up a squash commit dumping my current impressions.

I'd also like to use our `telephone` type here, although we'd want to 500 anybody setting `pref` and non-voice types.  In the absence of backend migrations adding _pref, etc., the API's `telephone` type would be even more aspirational here than it already is in the registration endpoint, but I think that's probably better than having multiple telephone types in the API.  Once we drop the old website, we can do a better job tracking telephone metadata like pref/type.  Alternatively, we can keep this flat and use the “E.123 telephone number” phrasing, with the understanding that we'll come back and give this a closer look after we drop the old website.

public.json
```diff
@@ -3183,6 +3183,79 @@
...
+              "retail-store",
+              "seasonal",
+              "wholesale-distributor"
+            ]
+          }
+        },
+        "buyer": {
+          "type": "string"
+        },
+        "contact-need": {
```
wking: Maybe `pickup-contact`?  There's a lot of contact information in these preferences, and anything that helps hint at this being for customers picking up orders would be useful.

public.json
```diff
@@ -3222,6 +3342,10 @@
           "type": "number",
           "format": "float"
         },
+        "order-minimum-adjustment": {
+          "type": "number",
+          "format": "float"
+        },
```
wking: This should not be exposed through the API until we are using it to manage drops internally.  As far as customers and drop coordinators are concerned, the drop [order-minimum][1] from 6a37b83c #26 is sufficient.  They don't need insight into our algorithm for calculating that number.

[1]: https://github.com/azurestandard/api-spec/blob/b693d1abe98c6e8b7c1d9180dae5f3bca79ec9bb/public.json#L3220

public.json
```diff
@@ -3198,9 +3304,23 @@
...
-            "public"
+            "public",
+            "other-building",
+            "other"
           ]
         },
+        "parking-other-notes": {
+          "description": "only available if parking type is \"other\"",
+          "type": "string"
+        },
```
wking: I'm on the fence about this.  Part of me wants to handle both the enum and other entries in the same field (e.g. support `parking = 'public'` and `parking = 'Some other value'`.  Looking through the database for retail drops turns up 254 active drops like this with 162 values:

    prefs = DropPointManagerPreferences.objects.filter(drop_point__active=True).exclude(other_parking=None).exclude(other_parking='')
    prefs.count()  # 254
    len(others)  # 162
    sorted(others)
    # [u'A truck stop/ travel ',
    #  u'AFF Dock',
    #  u'AFF warehouse',
    #  u'AML barge ',
    #  u'Abandoned grocery store',
    #  u'Alaska Air',
    #  u'Alco Store Parking Lot',
    #  …
    #  u'barge ',
    #  u'barge line Seattle, then I pickup at barge company in Ketchikan',
    #  u'behind Tractor Supply  Farm Store',
    #  u'business parking lot',
    #  …
    #  u'corner of Moran Drive & Wil Rogers',
    #  u'dirt pull off',
    #  u'dock for shipment to Guam',
    #  …
    #  u'undetermined',
    #  u'vacant building parking lot',
    #  u'vacant lot',
    #  u'vacant lot next to post office',
    #  u'warehouse']

My take on those entries is that they should have either used one of the non-other enum entries or that they're using this field for directions/location information (which should go in `DropPoint.directions`, although that may be intended for drivers, and not customers…).  Anyhow, the *other* approach has the benefit of matching our current database, but it's ugly enough that I think we want a clearer picture of what it's enabling before we bake it into the API.

public.json
```diff
@@ -3183,6 +3183,79 @@
...
+        "driver-callable-phone": {
+          "type": "string"
+        },
+        "eligibility-override": {
+          "type": "string",
+          "enum": [
+            "auto",
+            "force-off",
+            "force-on"
+          ]
```
wking: This is another internal field that we probably don't want to expose to customers/coordinators.

public.json
```diff
@@ -3198,9 +3304,23 @@
...
           ]
         },
+        "parking-other-notes": {
+          "description": "only available if parking type is \"other\"",
+          "type": "string"
+        },
+        "publishable": {
+          "description": "business preference",
+          "type": "boolean"
+        },
```
wking: I'm pretty sure we don't use this setting.  David said the original intention was something other than showing a drop on our drop-finder map, but I don't remember the exact details.  @tompetersjr, does that match what you remember from Monday meetings?

tompetersjr: I cannot remember for sure but that sounds about right.

public.json
caseybessette: My digging around in the code/templates didn't turn up any hints about what this is used for.  It's editable by drop coordinators, and although the API spreadsheet says members may have read access to it, drop members in the old site cannot see it (at least, I couldn't find anywhere that they'd be able to). @tompetersjr Could you suggest someone I could talk to about finding more details on this, or would you mind doing it?  The basic question is: How is this data used and can we (1) turn it into a free text field or (2) see about doing away with the extra "other" data (reviewing what's already in there to create new options, etc.) or (3) is there a good reason to keep the current model with optional "notes" if other is selected?

public.json
caseybessette: A useful query for getting all the other notes:
```sql
WITH other AS (
SELECT other_business_location_type AS other
  FROM drops_droppoint droppoint
  JOIN drops_droppointbizmanagerpreferences businesspreference
    ON businesspreference.drop_point_id = droppoint.id
 WHERE droppoint.active
   AND businesspreference.other_business_location_type IS NOT NULL
   AND businesspreference.other_business_location_type != ''
UNION
SELECT other_parking AS other
  FROM drops_droppoint droppoint
  JOIN drops_droppointmanagerpreferences retailpreference
    ON retailpreference.drop_point_id = droppoint.id
 WHERE droppoint.active
   AND retailpreference.other_parking IS NOT NULL
   AND retailpreference.other_parking != ''
)
  SELECT trim(other)
    FROM other
GROUP BY trim(other)
ORDER BY trim(other);
```

public.json
caseybessette: A useful query for getting all the other notes:
```sql
WITH other AS (
SELECT other_business_location_type AS other
  FROM drops_droppoint droppoint
  JOIN drops_droppointbizmanagerpreferences businesspreference
    ON businesspreference.drop_point_id = droppoint.id
 WHERE droppoint.active
   AND businesspreference.other_business_location_type IS NOT NULL
   AND businesspreference.other_business_location_type != ''
UNION
SELECT other_parking AS other
  FROM drops_droppoint droppoint
  JOIN drops_droppointmanagerpreferences retailpreference
    ON retailpreference.drop_point_id = droppoint.id
 WHERE droppoint.active
   AND retailpreference.other_parking IS NOT NULL
   AND retailpreference.other_parking != ''
)
  SELECT trim(other)
    FROM other
GROUP BY trim(other)
ORDER BY trim(other);
```

wking: On Thu, Jan 21, 2016 at 09:16:25AM -0800, Casey M. Bessette wrote:
> >            ]
> >          },
> > +        "parking-other-notes": {
> > +          "description": "only available if parking type is \"other\"",
> > +          "type": "string"
> > +        },
> 
> My digging around in the code/templates didn't turn up any hints
> about what this is used for.

In an October Monday meeting, David said they were prompts to collect
information we might be interested in [1].  I'm not sure why we'd
collect parking info in particular without exposing the other values
to customers (it doesn't seem like customer service would be
interested, and going from the database to customers via
customer-service and a phone line sounds to awkward to be the
intention).  I'll dig around and see if I can find anything.

[1]: https://github.com/azurestandard/azure/wiki/Midweek-Software-Meeting-20151020#drop-coordinator-preferences

caseybessette:
@wking I just started work on removing some of the drop preferences from the API to line up with your suggestions on this api-spec PR.  Are we sure we don't want to expose these through the API?  If the goal is to eventually implement everything in the API so even customer service will end up using it, why not just include these fields now with the appropriate customer-service-only permissions?  (Excepting "publishable" which apparently is deprecated and can/should be removed?)

wking:
On Thu, Jan 21, 2016 at 09:54:14AM -0800, Casey M. Bessette wrote:
> Are we sure we don't want to expose these through the API?  If the
> goal is to eventually implement everything in the API so even
> customer service will end up using it, why not just include these
> fields now with the appropriate customer-service-only permissions?

Because “so even customer-service will end up using it” is such a
long-term goal that I'd rather not weigh down the API until it's more
likely to become a reality.  Who knows if our underlying model will
even look like that by the time we get around to porting the
customer-service UI.  But to avoid losing work, I'm fine with the
customer-service-only fields you've added sitting in a do-not-merge
PR/branch to be revived if/when we get there.

caseybessette:
Okay, that makes sense. I'll just remove the fields from the PR then. I don't think it's not enough to warrant saving in a separate branch.

public.json
wking: On Thu, Jan 21, 2016 at 09:28:13AM -0800, W. Trevor King wrote:
> Thu, Jan 21, 2016 at 09:16:25AM -0800, Casey M. Bessette:
> > >            ]
> > >          },
> > > +        "parking-other-notes": {
> > > +          "description": "only available if parking type is \"other\"",
> > > +          "type": "string"
> > > +        },
> > 
> > My digging around in the code/templates didn't turn up any hints
> > about what this is used for.
> 
> … I'm not sure why we'd collect parking info in particular without
> exposing the other values to customers (it doesn't seem like
> customer service would be interested, and going from the database to
> customers via customer-service and a phone line sounds to awkward to
> be the intention).

The spec §9.7.4.6 (Security for the Drop endpoint [1]) shows the
parking fields as readable to drop members.  So I'm not sure if/where
we show this to customers in the website UIs, but this is
drop-member-readable information.

[1]: https://docs.google.com/document/d/1dwayyitTQT5ma-UhFfkC16i6aVrJ2ArV-HnUdWyst10/edit#heading=h.nwlzviwxircz

wking: On Tue, Jan 26, 2016 at 02:36:26PM -0800, W. Trevor King wrote:
> The spec §9.7.4.6 (Security for the Drop endpoint [1]) shows the
> parking fields as readable to drop members.

And the security spreadsheet (which overlaps with the spec table but
is more canonical since we reviewed the spreadsheet in a
Monday meeting [1]) also has the parking fields as readable to drop
members [2].

Grepping the spec for “parking” doesn't turn up any hits that look
like “here's where we show drop members this information”.  If we were
going to show this information to drop members (and that seems
reasonable to me), it would get documented in §9.6.3 (Drop Details)
[3].

[1]: https://azurestandard.slack.com/archives/general/p1453848680000319
[2]: https://docs.google.com/spreadsheets/d/17UAPW96CtQZD4-oDFz8dt9qez6HEMH4gc6Lmy4iB-bs/edit
[3]: https://docs.google.com/document/d/1dwayyitTQT5ma-UhFfkC16i6aVrJ2ArV-HnUdWyst10/edit#heading=h.g9nmmflrplpz

wking: In the 2016-02-01 software meeting, we agreed to drop the other option from the API, and to expose parking in the drop-detail view on the website (for customers who have access to it).  So I think `parking-other-notes` should go away.  I think we should also drop the `other` enum entry from `parking`, because the distinction between “I haven't set this” and “I have checked, and none of your other choices match” seems unimportant for this field.  We'll be keeping those fields in Beehive for now (just in case ;), so we'll need to avoid setting `parking` when the backing Beehive value is either other or null, and we'll need special code to avoid clobbering the backing Beehive value on PUT.

caseybessette:
Remove the parking `other` enum and the `parking-other-notes` field. Rebased off master.

PR #63, opened 2016-01-21
Add audit piece end points and definitions

colin-merc-tech:


wking:
This looks pretty reasonable to me.  Some ideas:

* I'd like to avoid leaking the PieceMeta / Piece names out here,
  since everyone finds them confusing.  It's possible that the API's
  product / packaged-product terminology is equally confusing, but one
  set of confusing terminology in the API is better than two parallel
  sets ;).  How about ‘GET /audit/products’ and ‘POST
  /audit/packaged-product/{id}’?  I'm using POST instead of PUT
  because ‘GET /audit/packaged-product/{id}’ doesn't exist [1].  I'd
  also be fine adding the singular GET and using PUT.

* To make filtering easier, I'd suggest attaching the ‘audited’ to the
  PieceMeta.  I expect cases where someone starts to audit a PieceMeta
  but only has time to get through a subset of associated Pieces to be
  extremely rare / nonexistent.

* Instead of a boolean ‘is-audited’, I'd prefer a date-time (like we
  have now with PieceMeta.reviewed_at).  That approach makes it easier
  to setup a rolling audit if you need to make subsequent passes.
  With the POST approach, we don't have to reflect audited-ness in the
  pushed payload and can just set

    PieceMeta.piece_units_audited = datetime.datetime.now()

  whenever we receive any POST for an associated piece (which looks
  like what you're doing now by not PUTting is-audited, but you have
  Piece.units_audited or some such).

* Descriptions for the properties would be nice.  I'm not sure what
  unit / unit-other is about.  Is the former an enum?

[1]: http://tools.ietf.org/html/rfc7231#section-4.3.4

colin-merc-tech:
https://github.com/azurestandard/api-spec/commit/c17eae71ac1e4f617f9da83cd2dccb400880ded1 addresses your concerns above.

wking:
On Fri, Jan 22, 2016 at 06:46:31AM -0800, colin-merc-tech wrote:
> c17eae71 addresses your concerns above.

I've pushed 180aa882 on top of your branch with a few other changes
(described in the commit message, but comment inline if any of them
don't look right to you).  If that looks good, go ahead and
rebase/squash this back down to one commit with all the changes.

colin-merc-tech:
My only real concern is dropping unit_other, as that was explicitly noted in the requirements https://github.com/azurestandard/internal-website/issues/5 and will require some significant changes on both the back and front-end. @tompetersjr do you want to weigh in an let me know if you want us to spend more time on this to facilitate Trevor's request?

tompetersjr:
@wking:  I think we need unit_other as there are many units in Beehive that will not be in pint.  Even if we assume that if the unit is not in pint then it is an other this could create issue's as we implement ongoing changes in pint.  Can you give some more motivation as to why you think it should be removed and how to handle other?  I'm inclined to keep it.

wking:
On Mon, Jan 25, 2016 at 08:35:40AM -0800, Tom Peters wrote:
> I think we need unit_other as there are many units in Beehive that
> will not be in pint.

Yeah, I'm not suggesting we change that.  I'm just suggesting we only
use one field to pass the type through the API, since I don't see a
benefit to baking “known by Pint” into the API.

> Even if we assume that if the unit is not in pint then it is an
> other this could create issue's as we implement ongoing changes in
> pint.

I don't see a problem if we stick with unabbreviated Pint units.
There might be collisions in abbreviations (e.g. ‘pks’ for pecks or
packs [1]), but if the unit is ‘pecks’ then I think that's always
going to be clear regardless of whether our Pint installation
recognizes the name.

> Can you give some more motivation as to why you think it should be
> removed and how to handle other?

For motivation, it's because I don't see a benefit to distinguishing
between “known by Pint” and “a label for count” in client-side code.
Customers are just going to see the unit text, without any “this is
known by Pint” styling.  Product managers are going to be entering the
units (e.g. via this audit UI) that make the most sense to them, and
don't need to tell us whether or not they expect those units to be
known by Pint or not (we can just check when we recieve their POST).
If there's just one field to set/check, client-side code around units
will be simpler.

To @colin-merc-tech's point on changes [2], I don't expect
implementation to be that big a deal.  Minimal changes to the current
azurestandard/beehive#1362 (at azurestandard/beehive@faba5eb) would
have serialization logic (_piece_json) like:

  if (not piece.unit and not piece.unit_other) or not piece.packs:
      unit, packs = _parse_old_name(piece.name)
  else:
      unit = piece.unit
      if unit == 'other':
          unit = piece.unit_other

and deserialization logic (_new_piece_fields) like:

  data = {
      …
     'unit' = payload['unit']
  }
  try:
      Piece.unit_types.parse_expression(data['unit'])
  except pint.UndefinedUnitError:
      data['unit_other'] = data['unit']
      data['unit'] = 'other'

And to @colin-merc-tech's point on specs
(azurestandard/internal-website#5), yeah, sorry I missed the earlier.
I think the large number of ‘other’ settings in #61 made me question
what good they were.

[1]: https://github.com/azurestandard/beehive/issues/1364#issuecomment-174064727
[2]: https://github.com/azurestandard/api-spec/pull/63#issuecomment-174561633

colin-merc-tech:
@wking just to confirm I understand, you are saying that we should remove unit-other from the API but keep the fields on the model and just handle the loading/saving of those fields on the back-end?

> If there's just one field to set/check, client-side code around units will be simpler.

I think the select an option or write your own is a pretty uncomplicated and well known UI concept. As for client-side code, I can't imagine it really complicates things by any significance and the actual work to facilitate it is already complete.

@tompetersjr just let me know which way you would like to proceed and I'll take another run at the implementation branch to get it matching the changes. Then we can get the review process of that one started. Once Trevor has signed off on the spec, I'll have Josh start working on updating the front-end to match the accepted spec changes.

wking:
On Mon, Jan 25, 2016 at 01:01:31PM -0800, colin-merc-tech wrote:
> @wking just to confirm I understand, you are saying that we should
> remove unit-other from the API but keep the fields on the model and
> just handle the loading/saving of those fields on the back-end?

If I was writing the back-end from scratch, I'd probably go with a
single field.  But with the API using a single field, we can do
whatever we want in the backend and refactor it later if/when we have
time and decide a different way is better.  So I'm fine leaving the
backend split out (and “does Pint understand this” has a better case
as a backend distinction anyway).

> > If there's just one field to set/check, client-side code around
> > units will be simpler.
> 
> I think the select an option or write your own is a pretty
> uncomplicated and well known UI concept.

I'm not sure how Angular-compatbile combo-boxes are generally setup,
but yeah, it shouldn't be that different on the input side.  I think
it is easier on the output side (e.g. we'll eventually want to display
these units somewhere, and having only one field to look at there will
certainly be easier than wrapping each display in a filter or if/else.

> … and the actual work to facilitate it is already complete.

The audit workflow, yeah.  Is the customer-facing consumption code
done?

PR #66, opened 2016-01-27
Mt/add notifications

colin-merc-tech:
ref https://github.com/azurestandard/beehive/pull/1498

### DIFF
--- a/public.json
+++ b/public.json
@@ -299,6 +299,215 @@
         }
       }
     },
+    "/drop-memberships": {
+      "get": {
+        "summary": "Returns all drop memberships from the system that the user has access to",
+        "operationId": "findDropMemberships",
+        "tags": [
+          "drop-membership"
+        ],
+        "parameters": [
+          {
+            "name": "drop",
+            "in": "query",
+            "description": "drop IDs to filter by",
+            "required": false,
+            "type": "array",
+            "items": {
+              "type": "integer",
+              "format": "int64"
+            },
+            "collectionFormat": "csv"
+          },
+          {
+            "name": "filter-person",
+            "in": "query",
+            "description": "person IDs to filter by",
+            "required": false,
+            "type": "array",
+            "items": {
+              "type": "integer",
+              "format": "int64"
+            },
+            "collectionFormat": "csv"
+          },
+          {
+            "name": "limit",
+            "in": "query",
+            "description": "maximum number of results to return",
+            "required": false,
+            "type": "integer",
+            "format": "int32"
+          },
+          {
+            "name": "start",
+            "in": "query",
+            "description": "offset to the first result to return.  Use negative numbers to offset from the end of the result list.",
+            "required": false,
+            "type": "integer",
+            "format": "int32"
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "drop membership response",
+            "schema": {
+              "type": "array",
+              "items": {
+                "$ref": "#/definitions/dropMembership"
+              }
+            },
+            "headers": {
+              "Count": {
+                "description": "total number of matching results (how many you'd get if you didn't set `limit`)",
+                "type": "integer",
+                "format": "int32",
+                "minimum": 0
+              }
+            }
+          },
+          "default": {
+            "description": "unexpected error",
+            "schema": {
+              "$ref": "#/definitions/errorModel"
+            }
+          }
+        }
+      },
+      "post": {
+        "summary": "Create a new drop membership",
+        "operationId": "addDropMembership",
+        "tags": [
+          "drop-membership"
+        ],
+        "parameters": [
+          {
+            "name": "drop-membership",
+            "in": "body",
+            "description": "drop membership to add",
+            "required": true,
+            "schema": {
+              "$ref": "#/definitions/newDropMembership"
+            }
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "drop membership response",
+            "schema": {
+              "$ref": "#/definitions/dropMembership"
+            }
+          },
+          "default": {
+            "description": "unexpected error",
+            "schema": {
+              "$ref": "#/definitions/errorModel"
+            }
+          }
+        }
+      }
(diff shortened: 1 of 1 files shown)


wking:
On Wed, Jan 27, 2016 at 08:01:02AM -0800, colin-merc-tech wrote:
>   * Update person definitions to include new notification fields

I think the JSON would be easier to read if it looked like:

  "notifications": {
    "checkout": ["email"],
    "delivery": ["email", "text"],
    "payment": ["text"]
  }

That's the same content, so I don't expect it to be harder to
implement.  Clients that don't bake in slug → name stuff will be
easier to write, but I don't think we'll have UIs like that.  We could
have an API like:

  "notifications": {
    "checkout": {
      "description": "Sent whenever an order is checked out.",
      "channels": ["email"]
    },
    "delivery": {
      "description": "Sent whenever …",
      ["email", "text"],
    },
    …
  }

with the descriptions being coded into the backend, sent with outgoing
JSON, and ignored with incoming JSON.  But then you'd need backend
changes for customer-facing text changes, which we probably don't
want.

Is there a reason that a bunch of {event}-{channel}-notifications
booleans is easier to handle in the client than an object with
{event}: [{channel}, ...] entries?

colin-merc-tech:
> Is there a reason that a bunch of {event}-{channel}-notifications
booleans is easier to handle in the client than an object with
{event}: [{channel}, ...] entries?

I don't know about easier, I just followed the spec https://docs.google.com/document/d/1dwayyitTQT5ma-UhFfkC16i6aVrJ2ArV-HnUdWyst10/edit#heading=h.maqyyardcfix

colin-merc-tech:
> What's this about? I don't see any backing implementation in azurestandard/beehive#1498.

Are you referring to name? If so, the implementation is here: https://github.com/azurestandard/beehive/commit/5bd034bbeceb904b162e0217ca35755341c0ab8c#diff-64969757a8e9bc9605d59b0ce412b0bcR114

wking:
On Mon, Feb 01, 2016 at 11:00:01AM -0800, colin-merc-tech wrote:
> > What's this about? I don't see any backing implementation in
> > azurestandard/beehive#1498.
> 
> Are you referring to name? If so, the implementation is here…

Yeah, thanks for a link to the implementation.  It looks like that's
just going to give the D### drop code [1]:

  dpm = DropPointMembership.objects.get(customer__user__username='wking')
  str(dpm)  # 'D2715'

So I think we can drop dropMembership.name and have client code use
dropMembership.drop instead.

[1]: https://github.com/azurestandard/beehive/blob/5bd034bbeceb904b162e0217ca35755341c0ab8c/apps/drops/models.py#L668-L669

wking:
On Mon, Feb 01, 2016 at 10:58:32AM -0800, colin-merc-tech wrote:
> > Is there a reason that a bunch of {event}-{channel}-notifications
> > booleans is easier to handle in the client than an object with
> > {event}: [{channel}, ...] entries?
> 
> I don't know about easier, I just followed the spec [§9.4.1.5
> Additional Implementation Notes].

I think those spec notes are for the Beehive-implementation and not
the JSON API.  I think a ‘notifications’ object with ‘{event}:
[{channel}, ...]’ entries would be cleaner (only one place to document
the event itself, only one place to document notification channels,
only one notification key in the root customer object, …).  But the
current approach here is functionally equivalent, so I'm ok landing it
if you don't want to reroll.

colin-merc-tech:
> I think those spec notes are for the Beehive-implementation and not
the JSON API.  I think a ‘notifications’ object with ‘{event}:
[{channel}, ...]’ entries would be cleaner (only one place to document
the event itself, only one place to document notification channels,
only one notification key in the root customer object, …).  But the
current approach here is functionally equivalent, so I'm ok landing it
if you don't want to reroll.

Ah, you just mean in the json response? I don't mind rolling it up whatever way you want. Shouldn't be a big deal to change it in that respect.

wking:
On Mon, Feb 01, 2016 at 12:25:24PM -0800, colin-merc-tech wrote:
> Ah, you just mean in the json response? I don't mind rolling it up
> whatever way you want. Shouldn't be a big deal to change it in that
> respect.

Yeah, just the JSON.

colin-merc-tech:
@wking  293ef18 fixes everything but the notification field json formatting and replacing name and drop id with what I'm assuming will be an object. I'm going to get you to go ahead and make those two changes to your liking so we can avoid batting it back and forth.

colin-merc-tech:
Just took a stab at it in the interest of keeping this moving along. See above commits and make any changes/fixes you require.

wking:
On Thu, Feb 04, 2016 at 10:48:25AM -0800, colin-merc-tech wrote:
> Just took a stab at it in the interest of keeping this moving
> along. See above commits and make any changes/fixes you require.

I just pushed 4aee397 to factor out the channel choices.  Except for
person.invoice (you'd repeated ‘payment’, but let me know if you have
wording you like better than ‘invoice’), which keeps its special
email/paper.  If we take that commit, drop d6d0c2bb [1], and squash
things up, I think this is good to go.  I'd like some more feedback on
d6d0c2bb though, since I don't want to drop something just because I
don't understand it ;).

[1]: https://github.com/azurestandard/api-spec/commit/d6d0c2bb9f3958cbf9226fa97cc4798e4e17cfb2#commitcomment-15928653

colin-merc-tech:
> I just pushed 4aee397 to factor out the channel choices.  Except for
person.invoice (you'd repeated ‘payment’, but let me know if you have
wording you like better than ‘invoice’), which keeps its special
email/paper.  If we take that commit, drop d6d0c2b [1], and squash
things up, I think this is good to go.  I'd like some more feedback on
d6d0c2b though, since I don't want to drop something just because I
don't understand it ;).

4aee397 looks good and I'm okay with dropping d6d0c2b. d6d0c2b was just the result of me misunderstanding both how a "drop_code" is generated and what you were requesting for the spec. I'm good with this if you're ready to merge.

PR #67, opened 2016-02-02
move audited attribute from productAudit to packagedProductAudit

MercenaryJosh:
The size, unit, and packs are what is being audited and those are
attributes of the packaged-product, therefore the audited attribute
should be on the packaged-product, not the product.

### DIFF
--- a/public.json
+++ b/public.json
@@ -1197,7 +1197,7 @@
           {
             "name": "audited",
             "in": "query",
-            "description": "filter for products that have been audited or not audited",
+            "description": "filter for products that have or haven't been audited.  When set, only products that contain (un)audited packaging are returned, and their packaging array only contains (un)audited packaging.",
             "required": false,
             "type": "string"
           },
@@ -5326,10 +5326,6 @@
           "description": "sentence-length product name.  Copied from `#definitions/product` for easier UI development",
           "type": "string"
         },
-        "audited": {
-          "description": "a datetime denoting if and when the product was last audited",
-          "type": "date-time"
-        },
         "packaging": {
           "description": "all related audited and/or non-audited packaged versions of this product",
           "type": "array",
@@ -5364,6 +5360,10 @@
         "unit": {
           "description": "the new unit of measurement.  You can use any string, but the backend will only be able to make automatic comparisons if you use SI prefixes (e.g. \"mm\" for millimeters) and stick to the following base units: pound, ounce, gram, gallon, liter, quart, pint, cup, fluid ounce, yard, and count",
           "type": "string"
+        },
+        "audited": {
+          "description": "a datetime denoting if and when the packaged-product was last audited",
+          "type": "date-time"
         }
       },
       "required": [


wking:
On Tue, Feb 02, 2016 at 09:19:23AM -0800, MercenaryJosh wrote:
> The size, unit, and packs are what is being audited and those are
> attributes of the packaged-product, therefore the audited attribute
> should be on the packaged-product, not the product.

The rationale for attaching it to the product was that auditing all
packaging associated with a product will likely happen in a very short
time-span, and filtering with /audit/products?audited=… is slightly
more complicated when products can contain both audited and
not-audited packaging [1].

But productAudit.packaging still has the “audited and/or non-audited
packaged versions” wording, so moving back to productAudit should be
fairly straigtforward.  This probably also needs an update to the
description for /audit/products audited parameter explaining how
filtering is going to work when a product is partially audited.  Maybe
just spell it out with something like:

  filter for products that have or haven't been audited.  When set,
  only products that contain (un)audited packaging are returned, and
  their packaging array only contains (un)audited packaging.

[1]: https://github.com/azurestandard/api-spec/pull/63#issuecomment-173628653

MercenaryJosh:
As is, there is no way for the front end to determine which packaged-products have been audited and which haven't. By putting the audited flag on the product, that would indicate that **all** of the packaged-products for that product have been audited. So then why post to /audit/packaged-product? Just make the post /audit/product and send all the packaging as audited? That would force the user to audit all the packaged-products before submission instead of relying on an abstract "will likely happen" and leaving open the possibility of inconsistent data.

wking:
On Fri, Feb 05, 2016 at 05:36:42PM -0800, MercenaryJosh wrote:
> As is, there is no way for the front end to determine which
> packaged-products have been audited and which haven't.

I don't see that as being a big deal.  The usual workflow would be:

1. Auditor audits a packaged-product (marking the whole product as
   audited, despite a few sizes not being checked yet.
2. Auditor audits remaining packaged-products, and now everything is
   peachy.

With the failure mode being:

1. Auditor audits a packaged-product (marking the whole product as
   audited, despite a few sizes not being checked yet.
2. Auditor goes away and forgets to ever finish auditing that product.

That could technically happen, but we can avoid it through training
(and this is a one-off UI, isn't it?).

If we do hit that failure mode (or for some other reason the auditor
pushes bogus data), we'll have bogus data in the database.  The
side-effect of bogus size data is that as-low-as pricing either fails
or gives weird results (depending on whether the bogus entries for a
product's packaging have similar types of units).  Since we're still
displaying per-package pricing, I think an extremely rare broken
as-low-as value won't be a big deal and can be corrected once someone
notices it.

> By putting the audited flag on the product, that would indicate that
> **all** of the packaged-products for that product have been
> audited. So then why post to /audit/packaged-product? Just make the
> post /audit/product and send all the packaging as audited?

That's fine with me too.  So possible approaches that I'm fine with
are:

a. Leave it as it stands, with the potential audit loophole of
   “auditor abandons audit part-way through a product's packaging” but
   simple implementation.

b. Land packagedProductAudit.audited like 0ecfe524, with the
   associated adjustment to the /audit/products?audited=… query
   parameter description [1].

c. Land a commit that removes ‘POST /audit/packaged-product/{code}’
   and replaces it with ‘POST /audit/product/{id}’.

I just don't want to land 0ecfe524 *without* updating the query
parameter description to explain how it handles partially-audited
packaging sets.

[1]: https://github.com/azurestandard/api-spec/pull/67#issuecomment-180630420

colin-merc-tech:
>b. Land packagedProductAudit.audited like 0ecfe52, with the
   associated adjustment to the /audit/products?audited=… query
   parameter description [1].

43eab20 addresses above

PR #68, opened 2016-02-05
public.json: Add volume and weight to packaged-product

heuscher:
For azurestandard/website#143, we'll need to calculate order-line weight and volume, without creating an order-line via the API.

### DIFF
--- a/public.json
+++ b/public.json
@@ -3838,6 +3838,16 @@
           "description": "size of the product (e.g. 12 oz., or 6x12 oz.)",
           "type": "string"
         },
+        "weight": {
+          "description": "weight of this packaged product in pounds (estimated until picking time)",
+          "type": "integer",
+          "format": "float"
+        },
+        "volume": {
+          "description": "volume of this packaged product in cubic feet",
+          "type": "integer",
+          "format": "float"
+        },
         "price": {
           "#ref": "#/definitions/price"
         },


public.json
```diff
@@ -3838,6 +3838,16 @@
           "description": "size of the product (e.g. 12 oz., or 6x12 oz.)",
           "type": "string"
         },
+        "weight": {
+          "description": "weight of this packaged product in pounds (estimated until picking time)",
```
wking: The current website displays a range of weights for variable-weight products.  We probably want to reflect that in the API, although it doesn't look like it's handled in [§13.7.3 (Product Items)][1] of the spec yet.  Here's what the current Beehive UI looks like:

variable-weight


[1]: https://docs.google.com/document/d/1dwayyitTQT5ma-UhFfkC16i6aVrJ2ArV-HnUdWyst10/edit#heading=h.h722kfhr0o9g

wking: azurestandard/website#159 has some thoughts on the UI for this, but [we're pretty sure][1] we'll need variable-weight information from the API.  To avoid recording variable-ness in a separate location, I'd recommend:

    "weight": {
        "minimum": ###,
        "maximum": ###,
        "average": ###,
    }

Folks who want to check variable-ness can compare `weight.minimum != weight.maximum`.  Our average weights are more guidelines than statistical measurements, so I wouldn't use anything as specific as “mean” or “mode” in the descriptions ;).

[1]: https://azurestandard.slack.com/archives/general/p1454710418001267

public.json
```diff
@@ -3838,6 +3838,16 @@
...
+          "type": "integer",
+          "format": "float"
+        },
+        "volume": {
+          "description": "volume of this packaged product in cubic feet (estimated until picking time)",
```
wking: We don't have variable-volume pieces, and the estimation for order-lines is mostly "we don't know how many we'll end up picking for your order", not "we don't know if we'll pick big ones or small ones".  So this should probably just stop after “… cubic feet”.

wking:
On Tue, Feb 16, 2016 at 09:29:19AM -0800, heuscher wrote:
> Merged #68.

Oops, I think this was back over to you to incorperate [1], see [2].
Can you file a follow-up PR adjusting the weight entry accordingly?

[1]: https://github.com/azurestandard/api-spec/pull/68#discussion_r52081306
[2]: https://azurestandard.slack.com/archives/general/p1455303730001905

PR #69, opened 2016-02-06
add endpoint for uploading a category image

MercenaryJosh:


wking:
On Fri, Feb 05, 2016 at 04:56:52PM -0800, MercenaryJosh wrote:
>   * add endpoint for uploading a category image

Instead of binding this to categories, I'd prefer an endpoint family
for CruD on files.  I don't know if that family needs a list, but at
least ‘POST /files’ and ‘DELETE /file/{id}’ would be good.  Files have
random paths by default [1], so I don't see a need for ‘UPDATE
/file/{id}’ (just DELETE the old and POST a new one).

Then models that reference files (e.g. categories, packaged-products,
etc.) can share the same file-management endpoints.  To associate a
file with a category (or packaged-product, etc.), users could use the
bare file ID in the JSON payload.  For example, POSTing:

  "image": "5882b5dc-2586-4766-ba06-bac9b4ca0332"

would attempt to attach the existing
File.get(path='5882b5dc-2586-4766-ba06-bac9b4ca0332') to the category
and return a response with:

  "image": "https://media.azurestandard.com/files/5882b5dc-2586-4766-ba06-bac9b4ca0332"

(or some such, depending on settings.MEDIA_URL).

How does that sound?

[1]: https://github.com/azurestandard/beehive/blob/e84a1370eda236140151a1a1c6b2721e5f9521d3/apps/files/models.py#L8

tompetersjr:
I'm torn on this one.

> Instead of binding this to categories, I'd prefer an endpoint family
for CruD on files.

On the one hand I see many uses for the flexibility for projects that we will be doing in the near future.  However we also need to get the website out as soon as possible.  I would say if retrofitting what you have to match a more generic spec is less then 4 hours work lets do it.  In that case it would be time well spent.  But if it is going to take more then that let's hold off for the time being.

wking:
On Sat, Feb 06, 2016 at 08:20:10AM -0800, Tom Peters wrote:
> On the one hand I see many uses for the flexibility for projects
> that we will be doing in the near future.

One example of where we would use a generic file create/delete
endpoint is the packaged-product image uploads we've been running
manually for azurestandard/beehive#667.  It would be nice to resolve
the image-maintenance portions of that and this category-image
requirement with the same endpoint.  As we roll out the new webiste,
getting faster turnaround on image uploads will become more important,
but I agree that it's not as critical as category images.

MercenaryJosh:
Instead of a generic file upload endpoint or a specific endpoint for only uploading an image, the image upload can be handled by the existing /categories and /category/{id} endpoints. I added the POST method to /categories and PUT method to /category/{id} endpoints since they were not present.

public.json
```diff
@@ -1911,6 +1911,42 @@
...
         }
+      },
+      "post": {
+        "summary": "Create a new category",
+        "operationId": "createCategory",
+        "tags": [
+          "category"
+        ],
+        "consumes": [
+          "multipart/form-data",
```
wking: Do we need to support form-data?  I'd rather keep input restricted to JSON if possible.  I'd expect the issue is accessing the image data to put into the JSON, and there are notes on that [here][1].

[1]: http://www.html5rocks.com/en/tutorials/file/dndfiles/

public.json
```diff
@@ -4553,12 +4634,9 @@
...
         },
-        "images": {
-          "type": "array",
-          "items": {
-            "description": "URL for a picture of this category",
-            "type": "string"
-          }
+        "image": {
+          "description": "URL for a picture of this category",
+          "type": "string"
```
wking: How does this work for POST/PUT?  This is currently a read-only field, and I think you need something like [this][1] to distinguish round-tripped URLs from new-image IDs.  I'd rather not have to look into the string to decide if it looks like a URL or JPEG (or whatever) to differentiate round-trips vs. new images (I'm happier looking into the string if it's [always a URL or portion thereof][1]).

[1]: https://github.com/azurestandard/api-spec/pull/69#issuecomment-180797097

public.json
MercenaryJosh: The main problem with using base64 is the size, but if that is not an issue, we can drop the multipart support and only use json.

public.json
MercenaryJosh: a quick implementation:

in _parse_category_json
````
try:
    payload['image'] = payload.get('image', '').decode('base64')
except:
    payload['image'] = None
````

in _update_category and _add_category
````
if payload['image']:
    category.image.save(str(uuid4()) + '.jpg', ContentFile(payload['image']))
````

File: public.json (hunk 1 above)
Comment (wking) [inline public.json]: I wouldn't worry about size, since we can replace this with a generic file-upload endpoint without too much trouble if that becomes a problem.

File: public.json (hunk 2 above)
Comment (wking) [inline public.json]: But if you're round-tripping a URL, it's not going to be a JPEG.

Comment (MercenaryJosh) [inline public.json]: If it is a round trip request, the image attribute on payload is going to get stripped out and replaced with None. Then the image attribute is not going to be updated, leaving the current value in place.

Comment (wking) [inline public.json]: Who strips it, the client?  That would be the first place (I think), where a GET/PUT round trip required client intervention to be a no-op.  And it means you couldn't remove an image from a category (you could only replace it with another image).

Comment (MercenaryJosh) [inline public.json]: When the payload is parsed in _parse_category_json. if the image url is present, the `.decode('base64')` will fail, and the catch block will set the `payload['image']` to None.

But that is not api-spec, it is implementation and thus outside the scope of this pull request. It was just an example of how the implementation could be done. (Though the implementation is, in fact, ready on the backend and frontend, and just waiting for this.)

Comment (wking) [inline public.json]: On Mon, Mar 14, 2016 at 06:07:34PM -0700, MercenaryJosh wrote:
> > +        "image": {
> > +          "description": "URL for a picture of this category",
> > +          "type": "string"
> 
> When the payload is parsed in _parse_category_json. if the image url
> is present, the `.decode('base64')` will fail, and the catch block
> will set the `payload['image']` to None.

Ah, thanks.  That's what I get for reviewing your comments on my
phone, which cutoff:

  payload['image'] = payload.get('image', '').decode('base64')

at:

  payload['image'] = payload.get('image', '')

> But that is not api-spec, it is implementation and thus outside the
> scope of this pull request. It was just an example of how the
> implementation could be done.

Knowing that you need to use base-64 to add a new image, and that
POST/PUTing anything which isn't valid base-64 (e.g. URLs with colons
and/or periods) is a no-op certainly sound like API-spec fodder to me.

How about a separate updateCategory definition with:

  "image": {
    "description": "A picture of this category.  Use base64 to upload new images, and leave unset to delete an existing image.  URLs containing colons and/or periods will result in no change.",
    "type": "string"
  }

That covers everything we can cover by my preferred generic endpoint
except for sharing a previously-uploaded image between categories and
products, etc.

Comment (wking) [inline public.json]: On Mon, Mar 14, 2016 at 09:30:02PM -0700, W. Trevor King wrote:
> "description": "A picture of this category.  Use base64 to upload
>    new images, and leave unset to delete an existing image.  URLs
>    containing colons and/or periods will result in no change.",

This doesn't clarify what happens for strings besides valid base64 or
URLs.  How about:

  "description": "A picture of this category.  Use base64 to upload
     new images, and leave unset to delete an existing image.  Invalid
     base64, including URLs containing colons and/or periods, will
     result in no change.",

File: public.json [hunk 3]
```diff
@@ -1945,6 +1977,47 @@
...
           }
         }
+      },
+      "put": {
+        "summary": "Update a category.",
+        "operationId": "updateCategory",
+        "tags": [
+          "category"
+        ],
+        "security": [],
```
wking: This endpoint needs to be secured (see security comments in 779331c6 (public.json: Add security and securityDefinitions fields, 2014-12-06).  `GET categories` is special because we don't do any auth for that endpoint (all the data is open).  Although we might want to double-check that since 9239898a (add "active" field to the category model, 2015-12-16, #54), and see if we're supposed to hide inactive categories from unauthenticated users, etc.

MercenaryJosh:
Looks good. So this should be ready.

PR #74, opened 2016-02-11
public.json: Fill in POST/PUT/DELETE for stop and support propagation

wking:
This requires backend tooling for azurestandard/internal-website#51
(as discussed with @tompetersjr and @jeffstrickler on 2016-02-10),
which allows for simpler client-side code.  Details in the commit
messages.

### DIFF
--- a/public.json
+++ b/public.json
@@ -1166,6 +1166,13 @@
             "schema": {
               "$ref": "#/definitions/updateStop"
             }
+          },
+          {
+            "name": "persist",
+            "in": "query",
+            "description": "if true, calculate the offset between the trips's delivery start and this stop's estimated time.  For any pre-cutoff trips associated with this stop's trip's route which cutoff later than this stop's trip, create a stop for this location (drop or pickup) if it does not already exist and update that stop's target and estimated times to have the same offset from that stops's trip's delivery start.  Also create a route-stop association, if it does not already exist, for this location on this stop's trip's route and update that route-stop with the same delivery offset.",
+            "required": false,
+            "type": "boolean"
           }
         ],
         "responses": {
@@ -1239,6 +1246,13 @@
             "schema": {
               "$ref": "#/definitions/updateStop"
             }
+          },
+          {
+            "name": "persist",
+            "in": "query",
+            "description": "if true, calculate the offset between the trips's delivery start and this stop's estimated time.  For any pre-cutoff trips associated with this stop's trip's route which cutoff later than this stop's trip, create a stop for this location (drop or pickup) if it does not already exist and update that stop's target and estimated times to have the same offset from that stops's trip's delivery start.  Also create a route-stop association, if it does not already exist, for this location on this stop's trip's route and update that route-stop with the same delivery offset.",
+            "required": false,
+            "type": "boolean"
           }
         ],
         "responses": {
@@ -1270,6 +1284,13 @@
             "required": true,
             "type": "integer",
             "format": "int64"
+          },
+          {
+            "name": "persist",
+            "in": "query",
+            "description": "if true, for any pre-cutoff trips associated with this stop's trip's route which cutoff later than this stop's trip, delete any stops for this location (drop or pickup).  Also delete any route-stop associations for this location on this stop's trip's route.",
+            "required": false,
+            "type": "boolean"
           }
         ],
         "responses": {
--- a/public.json
+++ b/public.json
@@ -1183,6 +1183,46 @@
             }
           }
         }
+      },
+      "put": {
+        "summary": "Update an existing stop",
+        "operationId": "updateStop",
+        "tags": [
+          "stop"
+        ],
+        "parameters": [
+          {
+            "name": "id",
+            "in": "path",
+            "description": "ID of stop to update",
+            "required": true,
+            "type": "integer",
+            "format": "int64"
+          },
+          {
+            "name": "stop",
+            "in": "body",
+            "description": "stop to update",
+            "required": true,
+            "schema": {
+              "$ref": "#/definitions/updateStop"
+            }
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "stop response",
+            "schema": {
+              "$ref": "#/definitions/stop"
+            }
+          },
+          "default": {
+            "description": "unexpected error",
+            "schema": {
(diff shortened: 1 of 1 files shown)


jeffstrickler:
I see a problem with the commits above. While django allows url parameters on PUT requests, my reading of the HTTP spec is that params should be in the body, just as in a POST request. Also, on DELETE - while it is permitted to add params to the url, it is generally considered bad form in part because of high profile cases where the google crawler has accidentally wiped out data by issuing a get that was treated as a delete - with all the params necessary in the url.

wking:
On Mon, Mar 07, 2016 at 05:28:10PM -0800, Jeff Strickler wrote:
> While django allows url parameters on PUT requests, my reading of
> the HTTP spec is that params should be in the body, just as in a
> POST request. Also, on DELETE - while it is permitted to add params
> to the url, it is generally considered bad form in part because of
> high profile cases where the google crawler has accidentally wiped
> out data by issuing a get that was treated as a delete - with all
> the params necessary in the url.

Can you point out the RFC sections you're referring to?  RFC 7231
§4.2.1 (Safe Methods) has [1]:

  When a resource is constructed such that parameters within the
  effective request URI have the effect of selecting an action, it is
  the resource owner's responsibility to ensure that the action is
  consistent with the request method semantics.  For example, it is
  common for Web-based content editing software to use actions within
  query parameters, such as "page?do=delete".  If the purpose of such
  a resource is to perform an unsafe action, then the resource owner
  MUST disable or disallow that action when it is accessed using a
  safe request method.  Failure to do so will result in unfortunate
  side effects when automated processes perform a GET on every URI
  reference for the sake of link maintenance, pre-fetching, building a
  search index, etc.

but we're just adjusting the scope of the action, not changing the
action itself.  That's the only “query” reference in RFC 7231 that
looks pertinent to me.

RFC 7230 §2.7.1 (http URI Scheme) has [2]:

  The hierarchical path component and optional query component serve
  as an identifier for a potential target resource within that origin
  server's name space.

RFC 7230 §A.2 (Changes from RFC 2616) has [3]:

  The segment + query components of RFC 3986 have been used to define
  the request-target, instead of abs_path from RFC 1808.

And RFC 3986 §3.4 (Query) has [4]:

  The query component contains non-hierarchical data that, along with
  data in the path component (Section 3.3), serves to identify a
  resource within the scope of the URI's scheme and naming authority
  (if any).

So it seems valid to me, and I prefer the query parameter for the
smaller API footprint (as I mention in the 316cf784 commit message).

I don't mind going with the “additional paths” route, it's just a bit
more copy-pasting in Swagger.  How do /propagate/stops and
/propagate/stop/{id} sound?  It doesn't feel quite right to me, so let
me know if you have something else you'd prefer.  Or if the RFC quotes
above make the query-parameter approach sound more reasonable to you
;).

[1]: http://tools.ietf.org/html/rfc7231#section-4.2.1
[2]: http://tools.ietf.org/html/rfc7230#section-2.7.1
[3]: http://tools.ietf.org/html/rfc7230#appendix-A.2
[4]: http://tools.ietf.org/html/rfc3986#section-3.4

colin-merc-tech:
re-based and added https://github.com/azurestandard/api-spec/pull/74/commits/f60a435e97da432930f4e5719cd0cedc6f1d52af and https://github.com/azurestandard/api-spec/pull/74/commits/56f1e8540e16bf4c3755eef1e4681ff5a8d82d44. If you'd like to keep the persist flag for PUT and POST and then we'll drop that commit and file a git issue to get POST and PUT up to speed on the back-end.

https://github.com/azurestandard/api-spec/pull/74/commits/f60a435e97da432930f4e5719cd0cedc6f1d52af Provides a more accurate description around what is actually happening with propagate and also adds persist.

colin-merc-tech:
> “If persist is set in addition to this flag…” seems like a typo in persist's own docs. I'm still wrapping my head around the new docs here, so I don't have a concrete suggestion for a fix yet.
Yeah, that was a typo. I fixed it and removed the commit that was removing the POST/DELETE propagate query params.

wking:
I just pushed d7aa621 → 11ad8ac which pushed the propagate → persist rename down into the commit landing the duration-pushing persistence.  Then I split d7aa621 into two commits, one which added the propagate parameter (with the “later stops on this trip” semantics) and one with the delta-duration persistence change for PUT (which I want to ask logistics about).

wking:
The current spec wording ([§1.5.2.3.1 Bulk edits][1]) is:

> This modification will be applied to all stops subsequent to, and beginning with the one you've selected upon clicking Save.

The “modification will be applied” wording sounds like the delta-duration logic from 11ad8ac, so we'll go with that until we hear otherwise (although we might want to make the spec and/or UI more explicit to avoid surprising people).

[1]: https://docs.google.com/document/d/16pfnAJAgdGvWfP3MNDHCCAexw9hAWyCx8desEaS7Ex8/edit#heading=h.kj2bpaz1sg2p

wking:
I just pushed 11ad8ac → 099caca with more detailed wording along the lines of the 11ad8ac ideas.  I'll post a more detailed breakdown of the changes after meeting with @tompetersjr.

wking:
I just pushed 099caca → 90355a7 with a comma fix.  Summarizing the changes between the earlier 11ad8ac and the current 90355a7 (`git diff --word-diff=color 11ad8ac..90355a7` is reasonably readable):

* Talk about estimated times (since we always adjust them).  Mention target-time adjustments as a special case for pre-cutoff trips (where the estimated and target time must always match).
* Replace “future created stops” with explicit references to the route-stop association.
* In persistent PUTs, explicitly ignore trips and routes that the location is not already associated with (the old approach in f737ae8 explicitly asked for *creating* missing associations).
* Replace “add that calculated time difference” and similar with “shift”.

wking:
And I just pushed 90355a7 → 26a5630 restoring the “both persist and propagate” are set wording I'd removed in 099caca.  The full change since the last time @colin-merc-tech touched the branch is d7aa621 → 26a5630.

wking:
I just pushed 5e4259ca → 6303d016 which leaves the code alone but streamlines the Git history (so it looks like we knew where we were headed from the start ;).

wking:
I merged this because we've merged and deployed the implementation in
azurestandard/beehive#1595 and azurestandard/beehive#1776.

PR #77, opened 2016-02-16
public.json: Fix integer -> number type for float formats

wking:
Comply with [the OpenAPI (previously Swagger) spec][1], fixing bugs
from 3e3b6084 (public.json: Add /order-lines and /order-line/{id},
2014-12-02) and a5e85e83 (public.json: Add orderLine.volume,
2015-09-18, #30).

[1]: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#data-types

### DIFF
--- a/public.json
+++ b/public.json
@@ -4576,12 +4576,12 @@
         },
         "weight": {
           "description": "weight of the shipped products in pounds (estimated until picking time)",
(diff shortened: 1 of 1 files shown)


wking:
These changes are being handled in #76.

PR #78, opened 2016-02-16
Fix 'persion' typo and broken JSON from #71

wking:
Typos from #71.


caseybessette:
Updated this to remove the fix for what Tom [already did](https://github.com/azurestandard/api-spec/pull/79#issuecomment-187937927).  Rebased off master.

PR #80, opened 2016-03-03
public.json: Add pick-offset and similar to the route models

wking:
Catch the route models (route and updateRoute) up with d0b5d33f (Merge
remote-tracking branch 'origin/mt/update-trip-api', 2015-12-03, #44).
We need the offsets to seed new trips.

I've left out the *-end offsets out, because James only wanted the
-end times for rare adjustments.  I prefer calculating them from the
route-stop offsets, and everyone seemed to be on board with that for
most cases.  When creating new trips, the end dates should be
calculated using either

a. new trip's *-start date-time + largest current route-stop delivery
   offset.
b. new trip's *-start date-time + difference between *-start and *-end
   date-times for the most-future existing trip on the route.

Whether a client uses (a) or (b) doesn't matter to me.

I've added a "delivery" qualifier to the description for
delivery-offset, to make the distinction between drops and pickups
we've setup (temporarily?) in d0b5d33f.

I've gone with integer-day offsets to match our current backend
implementation, but all of these offsets are between a cutoff
date-time and another date-time (e.g. delivery-start is a date-time),
so we should likely be using:

    "type": "string",
    "format": "duration"

like we have for routeStop.delivery-offset since d2a35b36
(public.json: Add 'GET /route-stops' and /route-stop/{id} endpoints,
2015-05-07).  To get something usable out the door faster, I'll defer
discussion on the best duration type until after the new logistics UI
is deployed.

Spun off of azurestandard/internal-website#64.

### DIFF
--- a/public.json
+++ b/public.json
@@ -3958,8 +3958,23 @@
           "type": "integer",
           "format": "int32"
         },
+        "pick-offset": {
+          "description": "number of days after cutoff for picking orders",
+          "type": "integer",
+          "format": "int32"
+        },
         "delivery-offset": {
-          "description": "number of days after cutoff before the first stop",
+          "description": "number of days after cutoff for the first delivery stop",
+          "type": "integer",
+          "format": "int32"
+        },
+        "backhaul-offset": {
+          "description": "number of days after cutoff for the first pickup stop",
+          "type": "integer",
+          "format": "int32"
+        },
+        "warehouse-arrival-offset": {
+          "description": "number of days after cutoff for the truck arriving at the warehouse",
           "type": "integer",
           "format": "int32"
         },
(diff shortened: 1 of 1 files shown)


wking:
For example, [here's the current Beehive logic][1] and [here's the current internal-website logic][2].

[1]: https://github.com/azurestandard/beehive/blob/f1501e81f3ae21c1d0df2f1664f1be73d247cb78/apps/routes/templates/routes/schedule_edit.html#L186-L190
[2]: https://github.com/azurestandard/internal-website/commit/3a76ee36277eb19551564b62c14cf2c2a887e9d6#diff-ef58c8471ff927c74e1ef2c9f7e1c04aR277

wking:
This was discussed in the [2016-03-07 meeting][1], and the conclusion was both sets of logic are fine.  We don't make one-off offsets frequently (e.g. bumping a single trip because of a holiday), and are fine handling those using the new logic where necessary.

[1]: https://github.com/azurestandard/azure/wiki/Monday-Software-Meeting-20160307#add-pick-offset-and-similar-to-the-route-models

PR #82, opened 2016-03-10
Add drop.publishable to definition.

caseybessette:
This is drop.business_preferences.publishable.


wking:
On Thu, Mar 10, 2016 at 02:45:16PM -0800, Casey M. Bessette wrote:
> This is drop.business_preferences.publishable.

Cross-linking azurestandard/beehive#1604.

caseybessette:
This property should have been removed from the spec and is no longer going to be supported.

PR #83, opened 2016-03-17
public.json: Add ?accept=<mime> to GET /order/{id}

wking:
JavaScript makes it easy to set headers on an `XMLHttpRequest` or
similar, but it's hard to set headers like that in an `<a href="…">`
link (there is an advisory [`type` attribute][1], but it's just a hint,
not a suggestion for the `Accept` header that the agent should use when
following that link).  You may be able to work around this with some
draft [File API][3] [blob manipulation][2] or by [reading the data into
memory and building a][4] [data URI][5], but those are both more trouble
than getting a PDF or CSV order representation should require.

This commit adds a new `accept` query parameter, so clients that have
trouble setting an `Accept` header have an alternative channel for
supplying that information.

It also documents (with a [per-operation `produces`][6]) the supported
MIME types for the order response.  The PDF MIME type is from [RFC
3778][7] and the CSV is from [RFC 4180][8].

[1]: https://www.w3.org/TR/html5/links.html#attr-hyperlink-type
[2]: http://stackoverflow.com/a/35632133
[3]: https://w3c.github.io/FileAPI/#dfn-createObjectURL
[4]: http://stackoverflow.com/a/20361958
[5]: http://tools.ietf.org/html/rfc2397
[6]: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#fixed-fields-5
[7]: http://tools.ietf.org/html/rfc3778
[8]: http://tools.ietf.org/html/rfc4180

### DIFF
--- a/public.json
+++ b/public.json
@@ -2202,7 +2202,19 @@
             "required": true,
             "type": "integer",
             "format": "int64"
+          },
+          {
+            "name": "accept",
+            "in": "query",
+            "description": "override the HTTP Accept header to chose the response type.  Empty and unrecognized values will result in application/json.",
+            "required": false,
+            "type": "string"
           }
+        ],
+        "produces": [
+          "application/json",
+          "application/pdf",
+          "text/csv"
         ],
         "responses": {
           "200": {


public.json
```diff
@@ -2162,8 +2162,20 @@
             "required": true,
             "type": "integer",
             "format": "int64"
+          },
+          {
+            "name": "accept",
+            "in": "query",
+            "description": "override the HTTP Accept header to chose the response type.  Empty and unrecognized values will result in application/json.",
+            "required": true,
```
colin-merc-tech: Is this value really required? the description and the backing code would indicate it's not.

wking: I think it is necessary.  See "This commit adds a new accept query parameter..." in the commit message and PR summary.

wking: See also “… it's hard to set headers like that…” in the commit message and PR summary.

colin-merc-tech: I just meant should the spec have "required": true. Seems like the accept query parameter isn't a required parameter.

wking: On Fri, Apr 08, 2016 at 06:03:30AM -0700, colin-merc-tech wrote:
> > +            "name": "accept",
> > +            "in": "query",
> > +            "description": "override the HTTP Accept header to chose the response type.  Empty and unrecognized values will result in application/json.",
> > +            "required": true,
> 
> I just meant should the spec have "required": true. Seems like the
> accept query parameter isn't a required parameter.

Ahh.  That was just a sloppy copy/paste.  Fixed in 9b4ef39 → a98fd7d,
which also rebases onto the current master.

PR #86, opened 2016-03-21
modify location model returned from /drops/locations

joshuahefner:


wking:
On Mon, Mar 21, 2016 at 11:25:36AM -0700, MercenaryJosh wrote:
>   * remove name attribute from locations model

If we don't need this, I'm OK dropping it.  I haven't been able to dig
up previous discussion that clearly motivates its inclusion and think
they're just inherited from 30f00542 ( Adds new markers api at
/api/drops/markers..., 2015-06-13).

>   * rename 'label' attribute to 'exclusivity'

I went with the generic ‘label’ in b1ce2df3 (public.json: Add
/drops/locations, 2015-06-25) because clients shouldn't care that this
label is currently just the exclusivity.  Details in the commit
message and some discussion around [1].

[1]: https://azurestandard.slack.com/archives/general/p1435273358004036

MercenaryJosh:
the drop-finder currently loads the drops from algolia by only retrieving the fields it needs (id, geo, and exclusivity). renaming the 'label' field to 'exclusivity will match the current drop model and mean that the drop-finder can pull data from either location.

wking:
On Mon, Mar 21, 2016 at 02:43:36PM -0700, MercenaryJosh wrote:
> the drop-finder currently loads the drops from algolia by only
> retrieving the fields it needs (id, geo, and exclusivity). renaming
> the 'label' field to 'exclusivity will match the current drop model
> and mean that the drop-finder can pull data from either location.

The drop model and the location model are similar, but they don't need
to be identical.  ‘label’ allows us to extend this to cover other
categories without altering the API.  For example see ‘inactive-open’
in b1ce2df3 and ‘Yours’ in [1].  For a UI based on the drop model:

  var label = loc.exclusivity || loc.label;

is pretty easy to drop in.

[1]: https://azurestandard.slack.com/archives/general/p1435273358004036

wking:
I'm closing this based on my above comments.

PR #91, opened 2016-04-05
public.json: Add a payroll-deduction payment method

wking:
So employees can mark orders they want paid for out of their next
paycheck.

There is no newPayrollDeduction, because these methods are created via
the Beehive UI and not via the API (yet).  And employees will not be
creating these on their own behalf (while most customers create
credit-card methods by uploading their credit-card information on
their own behalf).

There's no special information associated with this method, but we'll
keep a person ID for symmetry with the credit-card type.

### DIFF
--- a/public.json
+++ b/public.json
@@ -5028,10 +5028,40 @@
         "expiration"
       ]
     },
+    "payrollDeduction": {
+      "description": "a payroll deduction (one of our supported payment methods)",
+      "type": "object",
+      "properties": {
+        "id": {
+          "type": "integer",
+          "format": "int64"
+        },
+        "type": {
+          "description": "string identifying this payment model",
+          "type": "string",
+          "enum": [
+            "payroll-deduction"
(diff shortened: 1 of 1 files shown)


wking:
There are more notes on payroll deductions in [this comment][1] and later discussion in that issue.

[1]: https://github.com/azurestandard/beehive/issues/1583#issuecomment-205977542

wking:
Rebased around #70 with ab3d2d0 → 9f8450f.

caseybessette:
What should I do with this PR now that we're not implementing this payment method?

wking:
On Thu, Apr 07, 2016 at 06:41:12AM -0700, Casey M. Bessette wrote:
> What should I do with this PR now that we're not implementing this
> payment method?

We'll still want it so folks can checkout orders with payroll
deductions.  @tompetersjr, is that part of beta-2?  Later?

PR #94, opened 2016-04-13
packagedProductAudit change code to id

MercenaryJosh:
Code is a nullable field and not all packaged-products have a code.
This prevents that packaged-product from being audited. Changing the
identifier to id allows all packaged-products to be audited.

see azurestandard/internal-website#79

### DIFF
--- a/public.json
+++ b/public.json
@@ -1285,7 +1285,7 @@
         }
       }
     },
-    "/audit/packaged-product/{code}": {
+    "/audit/packaged-product/{id}": {
       "post": {
         "summary": "Audit an existing packaged-product",
         "operationId": "updatePackagedProductAudit",
@@ -1295,9 +1295,9 @@
         ],
         "parameters": [
           {
-            "name": "code",
+            "name": "id",
             "in": "path",
-            "description": "code of packaged product being audited",
+            "description": "id of packaged product being audited",
             "required": true,
             "type": "integer",
             "format": "int64"
@@ -5710,8 +5710,8 @@
       "description": "an audited or auditable packaged-product",
       "type": "object",
       "properties": {
-        "code": {
-          "description": "the packaged-product code (not unique to the audit process)",
+        "id": {
+          "description": "the packaged-product id (not unique to the audit process)",
           "type": "string"
         },
         "size": {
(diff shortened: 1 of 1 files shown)


public.json
```diff
@@ -5710,8 +5710,8 @@
       "description": "an audited or auditable packaged-product",
       "type": "object",
       "properties": {
-        "code": {
-          "description": "the packaged-product code (not unique to the audit process)",
+        "id": {
+          "description": "the packaged-product id (not unique to the audit process)",
```
wking: This also needs a `code` → `id` update in the `required` list below.

wking:
On Wed, Apr 13, 2016 at 01:46:29PM -0700, MercenaryJosh wrote:
> Code is a nullable field and not all packaged-products have a code.
> This prevents that packaged-product from being audited. Changing the
> identifier to id allows all packaged-products to be audited.

It took me a bit to assemble the history here.  The 'code' approach
landed in 2e232540 (Add audit piece end points and definitions,
2016-01-21, #63).  An earlier version of that PR used 'id', but I
proposed the 'id' -> 'code' change to avoid leaking information about
Pieces in the API [1,2].  The API implementation landed in
azurestandard/beehive@8f32e4e7 ( apps/api/views/piece_audit.py: Add
/audit/products and /audit/packaged-product/{code}, 2015-08-19,
#1362).  In reviewing that, I pushed gently for Product.code instead
of Piece.code [3,4] but didn't check for code-less Pieces.  I still
think we may not want to audit pieces without associated products [3].
So my preference would be to:

* Leave the API spec alone.
* Update the API implementation to use Product.code.
* Not audit Pieces that have no Products.

If we *do* need to audit Pieces that have not Products, then this PR
seems like the way to go.

And the audit API is internal and fairly peripheral, so if leaking
Piece information there is easier than the hour or so it would take to
adjust the API implementation to use Product.code, I'm ok with landing
this PR.

[1]: https://github.com/azurestandard/api-spec/pull/63#issuecomment-174068944
[2]: https://github.com/azurestandard/api-spec/commit/180aa8826ca8f8df66ef81a5164a8e0f53989c21
[3]: https://github.com/azurestandard/beehive/pull/1362#discussion_r52397170
[4]: https://github.com/azurestandard/beehive/pull/1362#issuecomment-184531877

PR #99, opened 2016-04-18
Update /favorites endpoint to support inlined properties.

caseybessette:
Currently only supports `packaged-product`.

### DIFF
--- a/public.json
+++ b/public.json
@@ -3402,6 +3402,13 @@
             "type": "boolean"
           },
           {
+            "name": "inline",
+            "in": "query",
+            "description": "for convenience, include a full representation of the inlined property. Currently supports \"packaged-product\" and \"packaged-product.product\".",
+            "required": false,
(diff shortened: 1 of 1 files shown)


public.json
```diff
@@ -3402,6 +3402,13 @@
             "type": "boolean"
           },
           {
+            "name": "inline",
+            "in": "query",
+            "description": "for convenience, include a full representation of the inlined property, currently only supports \"packaged-product\"",
```
wking: nit: “… property, currently…” → “… property.  Currently …”

wking:
I'm idling on this until we sort out how we're going to handle product
inclusion (via inline=packaged-product.product?) [1].  We can have
that discussion here or in azurestandard/beehive#1715, whatever folks
want.

[1]: https://github.com/azurestandard/beehive/pull/1715#discussion_r60137477

PR #100, opened 2016-04-20
Added category parameter to packaged-product-tags. Resolves #97.

### LINKED ISSUE #97: Filtering packaged-product-tags by category
Is this officially supported?
http://demo-beehive.azurestandardtech.com:8000/api/packaged-product-tags?internal=false&category=characteristic

I didn't see it in the spec but it works and I want to use it for Shopping Filters.

davidmcatee-azure:


public.json
```diff
@@ -1714,6 +1714,13 @@
             "type": "boolean"
           },
           {
+            "name": "category",
+            "in": "query",
+            "description": "category to filter by",
+            "required": false,
+            "type": "string"
```
wking: It looks like this should be:

            "type": "array",
            "items": {
              "type": "string"
            },
            "collectionFormat": "csv"

(like [/products has][0], although those are different categories ;) since the backend is [splitting on commas][1].

[0]: https://github.com/azurestandard/api-spec/blob/d81d58bbd8727cf701abafbb00c35970605f275a/public.json#L1373-L1382
[1]: https://github.com/azurestandard/beehive/blob/a2c3dbf5fb494f2fcd9a530142f0008d9ef6ac77/apps/api/views/packagedproduct.py#L418-L420

davidmcatee-azure:
Very informative, thank you!

PR #106, opened 2016-04-27
public.json: Add .active to payment methods

wking:
This allows us to return historical payment methods (e.g. so customers
can see which method they used for a past payment), while still
allowing customers to distinguish between methods that can and cannot
be used for new payments.

The permissions around toggling .active will be complicated.  For
example, a customer should be allowed to change creditCard.active from
true to false if they want to disable a particular card.  In that
case, they should also be able to immediately change .active back to
true.  However, they should not be able to re-activate cards that have
expired, or re-activate payroll-deductions if they are no longer an
employee.  The backend may implement whatever restrictions it feels
are appropriate in this respect, and clients should be prepared to
handle assorted 400s when .active-changing requests violate a
backend-imposed restrictoon.

See also azurestandard/website#209 and azurestandard/beehive#1582.

### DIFF
--- a/public.json
+++ b/public.json
@@ -2541,6 +2541,13 @@
         ],
         "parameters": [
           {
+            "name": "active",
+            "in": "query",
+            "description": "only return (in)active payment methods",
+            "required": false,
+            "type": "boolean"
+          },
+          {
             "name": "filter-person",
             "in": "query",
             "description": "person IDs to filter by",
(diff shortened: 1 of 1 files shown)


public.json
```diff
@@ -5023,6 +5030,10 @@
             "credit-card"
           ]
         },
+        "active": {
+          "description": "whether or not this payment method available for further activity"
```
caseybessette: Need a comma here :)

PR #107, opened 2016-04-27
public.json: Inlines for product.packaging and packagedProduct.product

wking:
Details in the commit-messages (which also includes a drive-by fix for
product's required array).

This covers [this request](https://github.com/azurestandard/beehive/pull/1715#issuecomment-215227195) while setting up the product and
packaged-product endpoints to be a bit more structured and performant.

### DIFF
--- a/public.json
+++ b/public.json
@@ -2366,7 +2366,7 @@
           {
             "name": "inline",
             "in": "query",
-            "description": "for convenience, include a full representation of the inlined property. Currently supports \"brand\", \"packaging\", and descendants.",
+            "description": "for convenience, include a full representation of the inlined property. Currently supports the local \"description\" and \"favorites\" as well as \"brand\", \"packaging\", \"vendors\", and descendants.",
             "required": false,
             "type": "array",
             "items": {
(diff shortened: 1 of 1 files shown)


davidmcatee-azure:
Looks good. While I still like the direction this is going with the `?inline` feature, it turns out I didn't need it after all. After I got deeper in working on [fixing the website favorites page](https://github.com/azurestandard/website/issues/245), I discovered that the fastest way to get product information was from azure-angular-providers `AzureProduct`, which queries algolia for each packaged product code and caches the results. Since the page was already doing this anyway, and the product pages also use the same query, I decided to use this approach instead of inlining packaged-product in the call to /favorites. Sorry for having you go through the trouble of adding this!

wking:
On Wed, Apr 27, 2016 at 04:18:58PM -0700, David McAtee wrote:
> Looks good. While I still like the direction this is going with the
> `?inline` feature, it turns out I didn't need it after all…

Ah, no worries.  In that case I suggest we table this PR until we
decide that the performance benefits are worth the implementation time
(maybe 4 hours?).

wking:
Something like this landed with azurestandard/beehive#2391.  When I get some time, I'll revive this PR to cover that change.

wking:
Rebased onto master and updated to cover azurestandard/beehive#2391 with a037bce → f4e336b.

PR #110, opened 2016-05-05
Add GET /*?pickup=... filtering where we have ?drop=... filtering

wking:
We want to be able to discover these associations regardless of
whether goods are moving on or off the truck.  This is a follow-up to
#49.

The route-stops implementation is still in flight with
azurestandard/beehive@93c48a35 (apps/api/views/routestop.py: Add
/api/route-stops and /api/route-stop/{id}, 2015-05-11,
tk/trip-stop-api), so I expect that will remain unimplemented for a
while.  But the other two should be easy to add.

### DIFF
--- a/public.json
+++ b/public.json
@@ -874,6 +874,18 @@
             "collectionFormat": "csv"
           },
           {
+            "name": "pickup",
+            "in": "query",
+            "description": "pickup IDs to filter by",
+            "required": false,
+            "type": "array",
+            "items": {
+              "type": "integer",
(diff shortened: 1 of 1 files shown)


wking:
We want the /route-stops change here for the “Add a backhaul” default discussed in azurestandard/internal-website/issues/89 (point 12 and later comments).  This feature is probably low-priority, but I've given this PR medium priority because it is the first step in a series of changes needed to implement the feature (which will also need backend and frontend changes).

PR #112, opened 2016-05-06
add type to payment model

MercenaryJosh:
adds the type of payment to the payment model to allow for
displaying the payment type as described in azurestandard/website#209

### DIFF
--- a/public.json
+++ b/public.json
@@ -5724,6 +5724,10 @@
           "type": "integer",
           "format": "int64"
         },
+        "type": {
+          "description": "The type of payment used",
+          "type": "string",
+        },
         "amount": {
           "description": "Amount to charge (or which was charged) in dollars",
           "type": "number",


wking:
On Fri, May 06, 2016 at 02:49:32PM -0700, MercenaryJosh wrote:
> adds the type of payment to the payment model…

This information is already in payment.payment-method's type [1,2,3].
Maybe we want the ability to inline payment.payment-method?

[1]: https://github.com/azurestandard/api-spec/blob/24c2d896adfea4db5059a558a6db76c1948c3837/public.json#L5701-L5710
[2]: https://github.com/azurestandard/api-spec/blob/24c2d896adfea4db5059a558a6db76c1948c3837/public.json#L5559-L5567
[3]: https://github.com/azurestandard/api-spec/blob/24c2d896adfea4db5059a558a6db76c1948c3837/public.json#L5669-L5677

MercenaryJosh:
To use the payment-method's type requires a separate api-call and then requires reformatting the text on the front end ('credit-card' to 'Credit Card', etc.) these may be small actions but they are unnecessary actions as there is virtually no cost to adding two lines to the _order_json and sending the reader-friendly text with the order json.

tompetersjr:
Looking at #113 and this I prefer this method for it's simplicity.  I also think this is easier to read but that is probably subjective.  So I'm making a judgement call and merging this one in as the implementation.

wking:
On Fri, May 06, 2016 at 03:12:00PM -0700, MercenaryJosh wrote:
> To use the payment-method's type requires a separate api-call…

No separate API call needed with #113; that's what the ?inline=…
construct is designed to address.

> … and then requires reformatting the text on the front end
> ('credit-card' to 'Credit Card', etc.)…

And if it's the backend's job to de-slug the payment-method type, now
you still have to do that when you show users a list of
/payment-methods.  

This change leaves the paymentMethod types (currently creditCard and
payrollDeduction) without knowledge of the new, API-provided
un-slugged type (only known to the payment model).

It also obscures the single source of truth for that type information.
When someone PUTs to /order/{id}, the current API isn't clear about
whether payment.type is respected, or about what happens if it
disagrees with payment.payment-method.type [1].  This is something
that can be clarified in the property description, but it's
clarification that the inline approach doesn't need.

It also doesn't scale as folks caring about order instances decide
they also care about other payment-method properties.  For example, if
the last-four go back into the order-detail page, the inline approach
already covers you.  With this approach, you have to make another
one-off copy from the paymentMethod model to the payment model, or go
back to making a secondary /payment-method/{id} request.

So while a single field copy like this PR landed isn't a big deal, I
think it's a step down the wrong path when we already have an
?inline=… solution that avoids these difficulties.

[1] POST/PUT/DELETE aren't currently in the spec for /orders at all.
    I'll file a separate PR fixing that.

tompetersjr:
Good feedback.

> This change leaves the paymentMethod types (currently creditCard and
payrollDeduction) without knowledge of the new, API-provided
un-slugged type (only known to the payment model).

Can you clarify that a bit.  Having a hard time grasping what this is getting at.  I think I need an example.

> It also obscures the single source of truth for that type information.
When someone PUTs to /order/{id}, the current API isn't clear about
whether payment.type is respected, or about what happens if it
disagrees with payment.payment-method.type [1].  This is something
that can be clarified in the property description, but it's
clarification that the inline approach doesn't need.

Good point.  I think documentation to clarify would work just fine stating that this is a read only property and ignored for POST/PUT/DELETE.

> It also doesn't scale as folks caring about order instances decide
they also care about other payment-method properties.  For example, if
the last-four go back into the order-detail page, the inline approach
already covers you.  With this approach, you have to make another
one-off copy from the paymentMethod model to the payment model, or go
back to making a secondary /payment-method/{id} request.

Also a good and valid point.  For now I'm good with that.  If we end up adding to it then we can look at switching it up.

wking:
On Fri, May 06, 2016 at 04:10:52PM -0700, Tom Peters wrote:
> > This change leaves the paymentMethod types (currently creditCard
> > and payrollDeduction) without knowledge of the new, API-provided
> > un-slugged type (only known to the payment model).
>
> Can you clarify that a bit.  Having a hard time grasping what this
> is getting at.  I think I need an example.

The website's payment method listings [1,2] are currently using the
de-slug filter here [3].  That filter gets fed paymentMethod instances
[4].  This PR added a convenient way to fill in a human-readable
method name for past orders [5], but does not help with those
payment-method displays that are managing payment methods in
isolation.  That means you'll still need the logic in [3].  And in
that case, what's the benefit to the API backend providing an
unslugged name?

[1]: https://website.azurestandard.com/#/my-account/profile
[2]: https://website.azurestandard.com/#/checkout/payments
[3]: https://github.com/azurestandard/website/blob/v0.1.20/app/scripts/filters/payment-method-info.js#L27
     Credit-card methods will use the other branch of that filter.
     The highlighted line is for non-credit-card methods.
[4]: Because when you're picking a method to use for that order, or
     for a one-off payment, you haven't gotten an order with the
     payment model from the API yet.
[5]: https://website.azurestandard.com/#/my-account/orders

wking:
Putting the previous comment another way.  If we want the API to give
an unslugged type for a given payment-method, that information belongs
in the payment-method model, not in a model like ‘payment’ that links
to the payment-method model.  Because “only consumers that get to this
payment-method via ‘payment’ will care about this field” is unlikely
to be true forever, even if it is true now (which it isn't [1,2]).

[1]: https://website.azurestandard.com/#/my-account/profile
[2]: https://website.azurestandard.com/#/checkout/payments

PR #113, opened 2016-05-06
public.json: Add GET /orders?inline=... and /order/{id}?inline=...

wking:
For folks who don't want to make a separate request to get the
associated payment method.

I considered replacing order.payment with an ID and having inline
support for that too, but that would require splitting the "please
charge this order to ..." modeling from "we charged that order to ..."
modeling.  That's probably a worthwhile change to make, but I'm
punting on it for now.

This PR is an alternative to #112.

### DIFF
--- a/public.json
+++ b/public.json
@@ -2450,6 +2450,17 @@
             "collectionFormat": "csv"
           },
           {
+            "name": "inline",
+            "in": "query",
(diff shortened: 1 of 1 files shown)


tompetersjr:
Closing this one out in favor of #112.

PR #116, opened 2016-05-09
public.json: Add additional generic payment methods (cash-on-delivery, etc.)

### LINKED ISSUE #111: Allow Net Terms Customers to Checkout
Allow customers with Net Terms to Checkout.  This may already be in the API-SPEC so please close if it is.

wking:
Details in the commit messages.

Fixes #111.

### DIFF
--- a/public.json
+++ b/public.json
@@ -5666,8 +5666,8 @@
         "expiration"
       ]
     },
-    "payrollDeduction": {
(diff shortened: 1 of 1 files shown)


wking:
Don't merge this yet; I'm going to rebase in better docs.

wking:
Better docs landed in 8b71ddf → f47fbf9, so I think this is good to go.

PR #119, opened 2016-05-09
public.json: Add dropMembership.pending

### LINKED ISSUE #115: indicate if a dropMembership is pending approval
The dropMembership model needs to indicate if it is still awaiting approval from customer service. This is part of the [spec](https://docs.google.com/document/d/1dwayyitTQT5ma-UhFfkC16i6aVrJ2ArV-HnUdWyst10/edit#heading=h.4lzssgbotfpz):

> A status of pending needs to be placed on the customer’s drop membership association with the ability for a customer service representative to change that flag to approved. When this feature gets added into beehive we can set all existing drop members as approved (pending false) to start with. Any new member added to a drop is set as approved except if they come from the website as described here on a semi-open drop. Prospective members can retrieve information on a drop’s location, drop id and future stops, and they can place orders on those stops. Otherwise, prospective members should have the same access to drop data as non-members. The idea is that in most cases, customer-service will be able to get prospective members approved as full members by cutoff time, so we don’t want to hold-up order placement. But we also don’t want to leak private information to un-approved customers.

This can be done by either adding a 'pending' boolean property or possibly a 'status' enum string property if there is the potential for other status (besides 'pending' and 'active') to be used in the future.

wking:
Allow provisional memberships (“Please let me in”) for folks joining
semi-open or closed drops.  Drop coordinators and/or customer-service
can hit:

    GET /drop-memberships?pending=true&...

to see who has asked to join drops (or, with `&drop={drop.id}`, who has
asked to join their drop).  They can fetch all memberships, pending or
accepted, by not specifying the `pending` parameter.  And they can
fetch only accepted memberships with `pending=false`.

Drop coordinators and/or customer-service can also accept requests by
PUTing with pending false.  And they can reject requests by DELETE'ing
the membership ([we aren't currently doing anything special for
rejections on the backend][1]).

I haven't setup a default value in `updateDropMembership` because the
reasonable choice depends on the drop being joined.  Folks joining an
open drop should set pending false (and the backend might 400 requests
that don't, because likely nobody will be checking to approve those
requests).  Folks joining semi-open or closed drops should set pending
to true, and the backend will 400 them if they don't.

Fixes #115.
Related to azurestandard/beehive#1476, azurestandard/beehive#1761, azurestandard/website#118.

[1]: https://github.com/azurestandard/beehive/issues/1761#issuecomment-217992838

### DIFF
--- a/public.json
+++ b/public.json
@@ -332,6 +332,13 @@
             "collectionFormat": "csv"
           },
           {
+            "name": "pending",
+            "in": "query",
+            "description": "only return pending (true) or accepted (false) memberships",
+            "required": false,
+            "type": "boolean"
+          },
+          {
             "name": "inline",
             "in": "query",
             "description": "for convenience, include a full representation of the inlined property.  Currently supports \"customer\" and \"drop\".",
@@ -4514,6 +4521,10 @@
           "type": "integer",
           "format": "int64"
         },
+        "pending": {
+          "description": "whether the membership is pending (true) or accepted (false).",
+          "type": "boolean"
+        },
         "notifications": {
           "$ref": "#/definitions/dropMembershipNotifications"
         }
@@ -4522,6 +4533,7 @@
         "id",
         "customer",
         "drop",
+        "pending",
         "notifications"
       ]
     },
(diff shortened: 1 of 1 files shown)


MercenaryJosh:
> Folks joining an open drop should set pending true ... Folks joining semi-open or closed drops should set pending to false

This should be the other way around. Open drops should have pending automatically set to false and semi-open or closed should be set to true.

wking:
On Mon, May 09, 2016 at 02:48:14PM -0700, MercenaryJosh wrote:
> This should be the other way around…

Oops, thanks.  Fixed with d525b10 → d788ffb.

PR #120, opened 2016-05-16
public.json: Include breakdown ancestors in packagedProduct.stock

wking:
The previous spec wording was unclear, but the current site includes
breakdown ancestors, so we want to mimic that here [1].  You can
almost figure out the ancestor-included values from the direct values
(e.g. if we set 0 stock for the 6 x 14 and 14 oz packaging in the new
doc example), but the API doesn't clarify what can be broken down into
what.

[1]: https://github.com/azurestandard/beehive/issues/1780

### DIFF
--- a/public.json
+++ b/public.json
@@ -4932,7 +4932,7 @@
           "#ref": "#/definitions/price"
         },
         "stock": {
-          "description": "amount of stock available for purchase",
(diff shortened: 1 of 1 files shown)


wking:
Merging this myself, since it's just documenting azurestandard/beehive#1780.

PR #121, opened 2016-05-16
public.json: Allow order.trip for parcel-carrier orders

wking:
The old website exposes this information with text like:

> You can continue modifying this order until May 16, 2016 at 6 p.m.

This change exposes parcel-carrier cutoff information to customers, so they can figure out when orders will cutoff (last call for (un)placing orders on that trip).

This could also allow customers to place orders on future parcel-carrier trips, although the Angular UI isn't planning on supporting that, and the backend might 400 such attempts.

I'm leaving order.drop unset for parcel-carrier orders, because it doesn't add any useful information (order.address already has the delivery location).

Also change “UPS” to the more generic “parcel-carrier”.

### DIFF
--- a/public.json
+++ b/public.json
@@ -5384,12 +5384,12 @@
           "format": "date-time"
         },
         "target-delivery": {
-          "description": "planned delivery time (fixed before the trip starts, and unset for UPS orders)",
+          "description": "planned delivery time (fixed before the trip starts, and unset for parcel-carrier orders)",
           "type": "string",
           "format": "date-time"
         },
         "estimated-delivery": {
-          "description": "estimated delivery time (updated after the trip starts until the order is delivered, and unset for UPS orders)",
+          "description": "estimated delivery time (updated after the trip starts until the order is delivered, and unset for parcel-carrier orders)",
           "type": "string",
           "format": "date-time"
         },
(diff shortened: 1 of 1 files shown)


wking:
There's some existing discussion of this issue in azurestandard/website#142, see [here][1].  Since then, there has been some discussion in Hangouts, but I'm not sure if those resulted in linkable notes.

[1]: https://github.com/azurestandard/website/issues/142#issuecomment-192464878

MercenaryJosh:
Would the trip be set on all parcel-carrier orders? I think it would make sense not to assign the trip (at least on the frontend) until the parcel-carrier order is placed. That way the user doesn't have to worry about the order rolling over to the next trip each night.

wking:
On Mon, May 16, 2016 at 11:41:35AM -0700, MercenaryJosh wrote:
> Would the trip be set on all parcel-carrier orders?

Yeah, I was planning on setting it as soon as an address was
associated.

> I think it would make sense not to assign the trip (at least on the
> frontend) until the parcel-carrier order is placed.

That is do-able, but means you wouldn't get the countdown timer for
open order (not that that's a big deal).

> That way the user doesn't have to worry about the order rolling over
> to the next trip each night.

What is the worry?  This is how we handle truck and will-call orders,
and we don't expect customers to worry about them (or if we do, we
have bigger problems than fixing the issue for parcel carriers ;).

MercenaryJosh:
the workflow I am thinking of is that if I am shopping on a parcel-carrier user, will I see a countdown until 6pm that night? Then at 6 pm will I see a 'cutoff has passed' message and then I need to either manually roll the cart over or wait until the parcel-carrier trip is verified and the order is automatically rolled over. Though those can be handled on the front end and do not have much to do with this PR.

wking:
On Mon, May 16, 2016 at 02:52:12PM -0700, MercenaryJosh wrote:
> the workflow I am thinking of is that if I am shopping on a
> parcel-carrier user, will I see a countdown until 6pm that night?

Yup.

> Then at 6 pm will I see a 'cutoff has passed' message and then I
> need to either manually roll the cart over or wait until the
> parcel-carrier trip is verified and the order is automatically
> rolled over.

Yup.  Although the window between the 6pm cutoff and the ~7pm
verification isn't going to be very long.

> Though those can be handled on the front end and do not have much to
> do with this PR.

And bumping unplaced orders that are past cutoff to the next available
trip is a general issue for all orders, and not specific to
parcel-carrier orders.

PR #124, opened 2016-06-03
Add countries and country-regions endpoints and definitions.

caseybessette:
Related API changeset: https://github.com/azurestandard/beehive/pull/1836

### DIFF
--- a/public.json
+++ b/public.json
@@ -3216,6 +3216,119 @@
         }
       }
     },
+    "/countries": {
+      "get": {
+        "summary": "Returns all countries from the system",
+        "description": "Countries that are allowed to be used with addresses, ordered by name.",
+        "operationId": "findCountries",
+        "tags": [
+          "country"
+        ],
+        "parameters": [
+          {
+            "name": "limit",
+            "in": "query",
+            "description": "maximum number of results to return",
+            "required": false,
+            "type": "integer",
+            "format": "int32"
+          },
+          {
+            "name": "start",
+            "in": "query",
+            "description": "offset to the first result to return.  Use negative numbers to offset from the end of the result list.",
+            "required": false,
+            "type": "integer",
+            "format": "int32"
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "country response",
+            "schema": {
+              "type": "array",
+              "items": {
+                "$ref": "#/definitions/country"
+              }
+            },
+            "headers": {
+              "Count": {
+                "description": "total number of matching results (how many you'd get if you could set an infinite `limit`)",
+                "type": "integer",
+                "format": "int32",
+                "minimum": 0
+              }
+            }
+          },
+          "default": {
+            "description": "unexpected error",
+            "schema": {
+              "$ref": "#/definitions/errorModel"
+            }
+          }
+        }
+      }
+    },
+    "/regions": {
+      "get": {
+        "summary": "Returns all country regions from the system",
+        "description": "Regions that are allowed to be used with addresses in the United States and Canada, ordered alphabetically.  Other countries may use any string values for their regions.",
+        "operationId": "findRegions",
+        "tags": [
+          "region"
+        ],
+        "parameters": [
+          {
+            "name": "country",
+            "in": "query",
+            "description": "country to filter by",
+            "required": false,
+            "type": "string"
+          },
+          {
+            "name": "limit",
+            "in": "query",
+            "description": "maximum number of results to return",
+            "required": false,
+            "type": "integer",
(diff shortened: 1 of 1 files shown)


public.json
```diff
@@ -3216,6 +3216,119 @@
...
+          "default": {
+            "description": "unexpected error",
+            "schema": {
+              "$ref": "#/definitions/errorModel"
+            }
+          }
+        }
+      }
+    },
+    "/country-regions": {
```
wking: This can just be `regions`.  We don't disambiguate it in the address model, and I can't think of any other region models we would be adding down the line.

public.json
```diff
@@ -3216,6 +3216,119 @@
...
+      "get": {
+        "summary": "Returns all country regions from the system",
+        "description": "Regions that are allowed to be used with addresses.  Only the United States and Canada have regions supported by the system.",
```
wking: Hmm.  I see that the backend does not support region at all (there's no free-form database field) for other countries.  Since the spec requires region, we probably want to say here that other countries are free to use whatever string they like for their region.  And then we'd 500 anyone who actually tried to do that until the backend was migrated to get `BaseAddress.region` or similar.

public.json
```diff
@@ -3216,6 +3216,119 @@
...
+        "operationId": "findCountryRegions",
+        "tags": [
+          "country-region"
+        ],
+        "parameters": [
+          {
+            "name": "filter-country",
```
wking: We only annotate by person (and not as frequently as we once did), so the filter here should just be `country`.

public.json
```diff
@@ -6075,6 +6188,7 @@
           "type": "string"
         },
         "region": {
+          "description": "required for the United States and Canada, not allowed for other countries",
```
wking: I think that lack of support is a bug.  I'd drop this line and make the `get.description` update [suggested above][1].

[1]: https://github.com/azurestandard/api-spec/pull/124/files#r65777525

wking: Ah, no.  We need to relax the region requirement for other countries to cover [Belize and similar][1].  So keep this line, and just say “optional for other countries”.

[1]: https://github.com/azurestandard/beehive/issues/1539

public.json
```diff
@@ -6091,10 +6205,32 @@
...
         "country-name"
       ]
     },
+    "country": {
+      "description": "a country",
+      "type": "object",
+      "properties": {
+        "name": {
+          "description": "country name",
+          "type": "string"
```
wking: Since the API also supports [ISO codes and such][1], we might as well include those in this model too.

[1]: https://github.com/azurestandard/beehive/blob/38ca8df51112c8f1767137df4bfa6a4febe16221/apps/api/views/address.py#L74-L77

public.json
```diff
@@ -6091,10 +6205,32 @@
...
+        }
+      }
+    },
+    "countryRegion": {
+      "description": "a country region (US state or Canadian province)",
+      "type": "object",
+      "properties": {
+        "name": {
+          "description": "region name",
+          "type": "string"
```
wking: Since the API [supports abbreviations][1], we should include those in this mode too.  We should also include a country reference (using the ISO3 as a key?).

[1]: https://github.com/azurestandard/beehive/blob/38ca8df51112c8f1767137df4bfa6a4febe16221/apps/api/views/address.py#L94-L96

public.json
```diff
@@ -6091,10 +6205,32 @@
...
+        },
+        "country-name": {
+          "type": "string"
+        }
```
wking: Ah, I see you have the country here.  I think ISO3s are more stable than names, so I'd suggest using them.

public.json
```diff
@@ -3216,6 +3216,119 @@
         }
       }
     },
+    "/countries": {
+      "get": {
+        "summary": "Returns all countries from the system",
+        "description": "Countries that are allowed to be used with addresses.",
```
wking: This should also specify the result ordering, which should be alphabetical by name.

public.json
wking: This should also specify the result ordering, which should be alphabetical by name.

public.json
```diff
@@ -3216,6 +3216,119 @@
...
+              "$ref": "#/definitions/errorModel"
+            }
+          }
+        }
+      }
+    },
+    "/regions": {
+      "get": {
+        "summary": "Returns all country regions from the system",
+        "description": "Regions that are allowed to be used with addresses in the United States and Canada.  Other countries may use any string values for their regions.  Ordered alphabetically, except that US states are ordered before other US territories.",
```
wking: Why the special ordering between US states and territories?

PR #126, opened 2016-06-10
public.json: Adjust PUT /order/{id} to return a changedOrder model

wking:
And add `?save=false` support.  This lets folks check the side effects
of an order-save (e.g. maybe they adjusted the shipment from a drop to
parcel-carrier, and want to know if we'll still ship all of their
ordered products via parcel-carrier).

PUTting an order can also change fees and such, but those aren't listed in
`.changes` because clients can compare the resulting `.order` with a previous
retrieval.  Comparing order lines with that approach would require more
work up-front to collect the original data, and more post-PUT work to see
if any lines had been removed or changed.

We currently only record line-removals in `.changes`.  But with the
coming price overhaul, shipping costs are probably going to be
per-line (Slack 2016-03-15), in which case we'll probably grow a
reprice-order-line orderChange type, or some such.

### DIFF
--- a/public.json
+++ b/public.json
@@ -5584,7 +5584,7 @@
           "format": "int64"
         },
         "trip": {
-          "description": "trip delivering the order (automatically populated for for parcel-carrier orders)",
+          "description": "trip delivering the order (automatically populated for parcel-carrier orders)",
           "type": "integer",
           "format": "int64"
         },
--- a/public.json
+++ b/public.json
@@ -2547,7 +2547,7 @@
     },
     "/order/{id}": {
       "get": {
-        "summary": "Returns a order based on a single ID",
+        "summary": "Returns an order based on a single ID",
         "operationId": "findOrderById",
         "tags": [
           "order"
@@ -2711,7 +2711,7 @@
     },
     "/order-line/{id}": {
       "get": {
-        "summary": "Returns a order-line based on a single ID",
+        "summary": "Returns an order-line based on a single ID",
         "operationId": "findOrderLineById",
         "tags": [
           "order"
--- a/public.json
+++ b/public.json
@@ -2543,6 +2543,38 @@
             }
           }
         }
+      },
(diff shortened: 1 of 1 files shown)


public.json
```diff
@@ -5599,6 +5699,49 @@
...
+          "description": "before shipping, the customer can use this to associate a payment-method with the order.  After shipping, it contains information about the charged payment",
+          "$ref": "#/definitions/payment"
+        },
+        "drop": {
+          "description": "drop the order is destined for (unset for parcel-carrier orders)",
+          "type": "integer",
+          "format": "int64"
+        },
+        "trip": {
+          "description": "trip delivering the order (automatically populated for for parcel-carrier orders)",
```
caseybessette: Should the description read "(automatically populated for truck orders)"?

wking: On Fri, Jun 10, 2016 at 11:54:15AM -0700, Casey M. Bessette wrote:
> > +        "trip": {
> > +          "description": "trip delivering the order (automatically populated for for parcel-carrier orders)",
> 
> Should the description read "(automatically populated for truck orders)"?

This is just copy/pasted from the ‘order’ model [1].  The information
should be manually populated for truck orders, and is automatically
populated for parcel-carrier orders.

[1]: https://github.com/azurestandard/api-spec/pull/121/files#diff-380fafdca01526afaf5cbc26fb2f0983R5418

caseybessette: Right, okay.  There's still a typo here with "for" occurring twice in a row.

wking: On Fri, Jun 10, 2016 at 12:02:18PM -0700, Casey M. Bessette wrote:
> > +        "trip": {
> > +          "description": "trip delivering the order (automatically populated for for parcel-carrier orders)",
> 
> Right, okay.  There's still a typo here with "for" occurring twice in a row.

Oops :p.  Fixed in 177201c → ecce47b.

PR #127, opened 2016-06-20
public.json: Add updatePackagedProductAudit.size

wking:
To allow for non-size distinctions between packaging without having to
split them into separate product families.  The vinegar example is for
CO463, but this change will also be useful for variable-weight
packaging such as MT159 (6 x 6-10 lb turkey).

### DIFF
--- a/public.json
+++ b/public.json
@@ -7042,6 +7042,10 @@
           "description": "the new unit of measurement.  Ideally one of \"\", or \"\", but you can use another unit if none of those fit.",
           "type": "string"
         },
+        "size": {
+          "description": "override the auto-generated size (based on packs and unit) with a custom name.  This can be useful for adding packaging information (e.g. \"32 floz glass\" to distinguish glass from plastic vinegar bottles).",
+          "type": "string"
+        },
         "clear-audited": {
           "description": "If true, clear the audited date-time for this packaged-product.  If false (the default), set the audited date-time to the current time.",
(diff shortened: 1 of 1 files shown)


public.json
```diff
@@ -7042,6 +7042,10 @@
           "description": "the new unit of measurement.  Ideally one of \"\", or \"\", but you can use another unit if none of those fit.",
           "type": "string"
         },
+        "name": {
```
caseybessette: Should this be called `size` instead of `name`?

wking: On Wed, Jun 22, 2016 at 01:46:22PM -0700, Casey M. Bessette wrote:
> > @@ -7042,6 +7042,10 @@
> >            "description": "the new unit of measurement.  Ideally one of \"\", or \"\", but you can use another unit if none of those fit.",
> >            "type": "string"
> >          },
> > +        "name": {
> 
> Should this be called `size` instead of `name`?

Ah, probably.  Want to file a PR?

caseybessette: This hasn't been merged yet. :)

wking: On Wed, Jun 22, 2016 at 01:51:23PM -0700, Casey M. Bessette wrote:
> > +        "name": {
> 
> This hasn't been merged yet. :)

Ah, nice :).  Updated in 43aa683 → 94bd9e8, although I don't like
“size” as much as I did before this PR (details in the new commit
message).

PR #132, opened 2016-07-13
public.json: Add /products?detailed=true

wking:
To provide a single endpoint:

```
/products?detailed=true&packaged-product={code}
```

that will return all the interesting information about a product (including its in-house and discontinued pieces).  More details in azurestandard/azure-angular-providers@30f479fd (Merge remote-tracking branch 'origin/dm/fix-double-algolia-product-requests, 2016-07-12, azurestandard/azure-angular-providers#56).

### DIFF
--- a/public.json
+++ b/public.json
@@ -1512,6 +1512,13 @@
             "type": "string"
           },
           {
+            "name": "detailed",
+            "in": "query",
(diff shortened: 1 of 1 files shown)


wking:
Obsolete with azurestandard/beehive#2391 (this needs to get rerolled to to use `?inline=…`).

wking:
I've rerolled #107 to cover azurestandard/beehive#2391, so I'm closing this.

PR #133, opened 2016-07-13
public.json: Split order.fees into its own endpoint

wking:
This allows the backend to use more expensive fee-estimation logic (e.g. hitting parcel-carrier APIs) without bogging down order manipulations.  We initially [expected order-manipulation hitting fee recalculation to be rare][1], but it turns out to be [unacceptably slow][2], and [we don't need the fee information most of the time][3] (or it's ok if the fee update shows up asynchronously a second or two so after the change that triggered it).

[1]: https://github.com/azurestandard/website/issues/186#issuecomment-213072160
[2]: https://github.com/azurestandard/website/issues/552#issue-165217215
[3]: https://github.com/azurestandard/website/issues/552#issuecomment-232406614

### DIFF
--- a/public.json
+++ b/public.json
@@ -3069,6 +3069,103 @@
         }
       }
     },
+    "/order-fees": {
+      "get": {
+        "summary": "Returns all order fees from the system that the user has access to",
+        "operationId": "findOrderFees",
+        "tags": [
+          "order"
+        ],
+        "parameters": [
+          {
+            "name": "order",
+            "in": "query",
+            "description": "order IDs to filter by",
+            "required": false,
+            "type": "array",
+            "items": {
+              "type": "integer",
+              "format": "int64"
+            },
+            "collectionFormat": "csv"
+          },
+          {
+            "name": "limit",
+            "in": "query",
+            "description": "maximum number of results to return",
+            "required": false,
+            "type": "integer",
+            "format": "int32"
+          },
+          {
+            "name": "start",
+            "in": "query",
+            "description": "offset to the first result to return.  Use negative numbers to offset from the end of the result list.",
+            "required": false,
+            "type": "integer",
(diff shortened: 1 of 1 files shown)


davidmcatee-azure:
I have a question about something I just noticed. Currently in the fees that are returned, the "shipping" type fee doesn't have an id. Is that intentional?

screen shot 2016-07-13 at 2 26 58 pm

wking:
On Wed, Jul 13, 2016 at 02:29:03PM -0700, David McAtee wrote:
> Currently in the fees that are returned, the "shipping" type fee
> doesn't have an id. Is that intentional?

The shipping fee is currently stored as a UPSOrderShipment field [1].
I'd like to drop that field and use an OrderFee instance, but fixing
that isn't a current priority (unless the frontend starts needing to
refer to fees by ID ;).

[1]: https://github.com/azurestandard/beehive/blob/d853c7cd59bfa957b350c07cc9f6f02b0c228420/apps/api/views/order.py#L88-L93

wking:
A side-effect of this change is that, with order.fees removed, folks
using #126's PUT order/{id}?save=false won't have a way to tell if the
fees changed as a result of their update.  @davidmcatee-azure and I
think it would be nice to expose that information, so that a user
moving an order from a truck to parcel carrier could see:

  If you make this change, we'll drop $ORDER_LINES and change your
  fees to $NEW_FEES.

(they have the old fees from their last order/{id} response).  That
would probably help push folks towards joining/using drops (which
saves them on shipping and leaves them more money to buy Azure
products :).

To keep exposing fee-change information after this API change, we
could add a fee-change type to our orderChange model.  However, since
the frontend is not currently using ?save=false, @davidmcatee-azure
and I agree that we probably want to punt that change into the future.

PR #138, opened 2016-07-25
public.json: Add trip.notes

wking:
David Stelzer pointed out that the customer experience is nicer if we can tell them why their stop is late, instead of just that it is late.

### DIFF
--- a/public.json
+++ b/public.json
@@ -5332,6 +5332,10 @@
           "description": "estimated date-time for warehouse arrival",
           "type": "string",
           "format": "date-time"
+        },
+        "notes": {
+          "description": "free-form Markdown notes for additional trip information.  We expect this will mostly be for delay causes (e.g. \"High winds in Laramie\" or \"The truck broke down\")",
+          "type": "string"
         }
       },
       "required": [


public.json
```diff
@@ -5332,6 +5332,10 @@
           "description": "estimated date-time for warehouse arrival",
           "type": "string",
           "format": "date-time"
+        },
+        "notes": {
+          "description": "free-form Markdown notes for additional trip information.  We expect this will mostly be for delay causes (e.g. \"High winds in Laramie\" or \"The truck broke down\")",
```
tompetersjr: I'm ok with this but the implementation really only needs to support text right now.  No need for Markdown.  Is there a reason to specifically mention Markdown?

wking: On Mon, Jul 25, 2016 at 01:29:00PM -0700, Tom Peters wrote:
> > +        "notes": {
> > +          "description": "free-form Markdown notes for additional trip information.  We expect this will mostly be for delay causes (e.g. \"High winds in Laramie\" or \"The truck broke down\")",
> 
> I'm ok with this but the implementation really only needs to support
> text right now.  No need for Markdown.  Is there a reason to
> specifically mention Markdown?

Not really.  Maybe you want to bold something?  It's a superset of
text/plain, and the backend doesn't care, so Markdown or not is really
up to “do we need styling, links, lists, …?” and “how easy will it be
for the frontend to support editing/rendering?”.  I'm not clear on the
former, and think “just expose the raw Markdown” is good enough for
an initial implementation of the latter.

If we want to get fancier on the frontend, this looks like a
reasonable embedable editor [1], and this looks like a reasonable
renderer [2].

[1]: http://epiceditor.com/
[2]: https://github.com/evilstreak/markdown-js

PR #142, opened 2016-08-04
include the last four digits of the credit card used to make a payment

MercenaryJosh:
required by azurestandard/website#710

Only include the last-four if the payment method was 'Credit Card'.

### DIFF
--- a/public.json
+++ b/public.json
@@ -6524,6 +6524,10 @@
           "description": "The type of payment used",
           "type": "string"
         },
+        "last-four": {
+          "description": "The last four digits of the credit card used. Only included if the payment type is Credit Card",
+          "type": "string"
+        },
         "amount": {
           "description": "Amount to charge (or which was charged) in dollars",
           "type": "number",


wking:
Can we just inline the payment method based on a query parameter?

MercenaryJosh:
At least some credit card payments were not returning a payment method. The value for the last four comes from the `order.checkout_payment`, while the request for CustomerPaymentMethod returns no  results. (at least on some orders. I haven't checked them all).

wking:
Missing methods should no longer be an issue now that #106 has been implemented

wking:
On Thu, Aug 04, 2016 at 05:37:48PM -0700, W. Trevor King wrote:
> Missing methods should no longer be an issue now that #106 has been
> implemented

customers/migrations/0024_payment_method_active_init.py (from
azurestandard/beehive#1976, implementing #106) set
CustomerPaymentMethod.active, but did not create methods for old
cards.

  SavedCreditCard.objects.filter(customerpaymentmethod=None).count()  # 9638
  SavedCreditCard.objects.filter(customerpaymentmethod=None, visible_to_customer=True).count()  # 18

Looking at the most recent visible card:

  card = SavedCreditCard.objects.filter(customerpaymentmethod=None, visible_to_customer=True).order_by('-created_at')[0]
  card.created_at.isoformat()  # '2016-03-25T09:33:13.286146'
  card.updated_at.isoformat()  # '2016-03-25T09:33:13.320625'
  [(ppc.id, ppc.created_at.isoformat()) for ppc in card.paymentprocessorcommunication_set.all()]
  # [(911635, '2016-03-25T09:46:37.683271'),
  #  (911630, '2016-03-25T09:37:48.285784'),
  #  (911623, '2016-03-25T09:33:15.353536'),
  #  (911622, '2016-03-25T09:33:13.273725')]

March was a long time ago, so I'll guess that this hole in
payment-method creation has been patched in the meantime.
Creating the missing methods:

  cc = PaymentTermType.objects.get(payment_code='CC')
  for card in SavedCreditCard.objects.filter(customerpaymentmethod=None):
      active = card.visible_to_customer and card.usable()
      CustomerPaymentMethod.objects.create(
          customer=card.customer,
          payment_term=cc,
          credit_card=card,
          active=active)

After this change in production, the order without a method [1] has
one:

  $ curl -su wking:… https://api.azurestandard.com/order/3796168 | jq '.["checkout-payment"]'
  {
    "last-four": "8912",
    "payment-method": 600939,
    "type": "Credit Card",
    "amount": 33.91,
    "paid": true
  }

And the lack of access to inactive payment methods [2,3] is resolved
by azurestandard/beehive#2046.  I realize that this has already
landed, but I still think that copying over payment-method attributes
one at a time into the payment model is unnecessarily ad-hoc.  From
[4]:

  It also doesn't scale as folks caring about order instances decide
  they also care about other payment-method properties.  For example,
  if the last-four go back into the order-detail page, the inline
  approach already covers you.  With this approach, you have to make
  another one-off copy from the paymentMethod model to the payment
  model, or go back to making a secondary /payment-method/{id}
  request.

And we're using the ?inline=… approach successfully in other
endpoints.

[1]: https://github.com/azurestandard/website/issues/209#issuecomment-198572279
[2]: https://github.com/azurestandard/website/issues/209#issuecomment-198583844
[3]: https://github.com/azurestandard/website/issues/209#issuecomment-199413748
[4]: https://github.com/azurestandard/api-spec/pull/112#issuecomment-217583186

caseybessette:
@MercenaryJosh If you want to work up an api-spec PR and a new Beehive PR, I will be happy to review them.  Trevor makes some good points here.  I probably shouldn't be reviewing PRs past my bedtime.  There are a lot of examples in the API code already for inlining models.

PR #145, opened 2016-08-10
public.json: Add orderLine.warnings

wking:
So we can return should-not-buy warnings like:

> {product.code} is likely to expire before it gets to {drop.name}
> ({drop.id}) via {route.code}.

This will help with cart rendering and quick-add.  More discussion
[here][1] and [here][2].

[1]: https://github.com/azurestandard/website/issues/754#issuecomment-238982722
[2]: https://github.com/azurestandard/website/issues/718#issuecomment-237717173


davidmcatee-azure:
This looks good. Can you create the beehive issue too?

PR #146, opened 2016-08-11
Add total-cases to order definition.

caseybessette:
Fixes #2073.

### DIFF
--- a/public.json
+++ b/public.json
@@ -6056,6 +6056,11 @@
         "notes": {
           "description": "Warehouse packing instructions",
           "type": "string"
+        },
+        "cases-shipped": {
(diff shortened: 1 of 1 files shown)


public.json
```diff
@@ -6056,6 +6056,11 @@
         "notes": {
           "description": "Warehouse packing instructions",
           "type": "string"
+        },
+        "total-cases": {
```
wking: This is similar to the existing `quantity-shipped`.  Can we put it right after `quantity-shipped` in the spec and call it `cases-shipped`?

And probably use an `int32` to match the `quantity-*` settings, although I doubt that distinction matters ;).

PR #147, opened 2016-08-16
public.json: Split out email addresses into their own endpoints

wking:
This gives us a convenient place to store errors (e.g. per-address bounce
reports) and lets us access multiple addresses per person
(although the backend currently caps that at three emails per person).
Multiple emails per person allows you to include others (e.g. family
members) in cutoff-notifcations and the like, although only the primary
address (with the highest preference) is used for password-reset, login,
and the like.

Also fill in the address API to catch up with the current
implementation, because I'm using the address API to template the new
email API.

Related issues: azurestandard/beehive#510, azurestandard/beehive#1806, azurestandard/website#624.

### DIFF
--- a/public.json
+++ b/public.json
@@ -3687,6 +3687,38 @@
             }
           }
         }
+      },
+      "post": {
+        "summary": "Create a new address",
+        "operationId": "addAddress",
+        "tags": [
+          "address"
+        ],
+        "parameters": [
+          {
+            "name": "address",
+            "in": "body",
+            "description": "address to add",
+            "required": true,
+            "schema": {
+              "$ref": "#/definitions/address"
+            }
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "address response",
+            "schema": {
+              "$ref": "#/definitions/address"
+            }
+          },
+          "default": {
+            "description": "unexpected error",
+            "schema": {
+              "$ref": "#/definitions/errorModel"
+            }
+          }
+        }
       }
     },
     "/address/{id}": {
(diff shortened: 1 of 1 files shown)


wking:
For folks who are interested, GitHub has a single [`/emails` endpoint][1], but the one I'm adding here is a bit different:

* It explicitly describes the confirmation endpoint, and allows a configurable base URL for folks who want to build alternative frontends.
* It uses an `error` string instead of a `verified` boolean, so the user can get more information about why we don't like a particular address.
* It uses an integer `preference` instead of a `primary` boolean, so users can order their whole set of emails, instead of just picking the first.  No real reason for this besides matching the address API, where ordering was a requested feature.

[1]: https://developer.github.com/v3/users/emails/

public.json
```diff
@@ -3835,6 +3935,239 @@
...
+        "summary": "Complete an in-progress email confirmation.",
+        "operationId": "emailConfirmation",
+        "tags": [
+          "email"
+        ],
+        "parameters": [
+          {
+            "name": "token",
+            "in": "body",
+            "description": "The confirmation token from ",
```
davidmcatee-azure: The confirmation token from ... ?

public.json
```diff
@@ -3835,6 +3935,239 @@
...
+            "required": true,
+            "type": "object",
+            "schema": {
+              "$ref": "#/definitions/passwordResetConfirmation"
+            }
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "person response",
```
davidmcatee-azure: Did you mean for this to return a person?

wking:
I just pushed b604752 → 3ab5c36 fixing:

* An incomplete description [1], and
* The wrong return type [2].

for POST /email/{id}/confirm.

[1]: https://github.com/azurestandard/api-spec/pull/147/files/b6047520d20eb8499214f12757032fad874902f4#r74888127
[2]: https://github.com/azurestandard/api-spec/pull/147/files/b6047520d20eb8499214f12757032fad874902f4#r74888280

davidmcatee-azure:
Looks good.

PR #148, opened 2016-08-16
include product id in packagedProductProduct

MercenaryJosh:
This is needed for the internal-website category app to link to
products on the new website. see azurestandard/internal-website#145

### DIFF
--- a/public.json
+++ b/public.json
@@ -5961,9 +5961,13 @@
       ]
     },
     "packagedProductProduct": {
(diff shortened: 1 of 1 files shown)


public.json
```diff
@@ -5961,9 +5961,13 @@
       ]
     },
     "packagedProductProduct": {
-      "description": "the name and brand of a product that a packaged-product is associated with",
+      "description": "the id, name, and brand of a product that a packaged-product is associated with",
```
wking: `id` → `ID`

PR #149, opened 2016-08-18
public.json: Add /emails/resend and adjust email requirements

wking:
Return resend tokens from POST and PUT, and add an `/emails/resend` endpoint to receive those tokens.  This allows the following workflow:

1. User submits new / updated email.

2. Frontend shows user:

   > You've submitted your update and will receive a confirmation
   >  email shortly.  You can click here to resend the email.

   POSTing the resend token to `/emails/resend` if the user clicks.

3. User gets the confirmation email and visits the confirmation URL.

4. Frontend gets the confirmation token from the URL, and submits
   token to `/emails/confirm`.  On success, frontend shows the user:

   > Email update confirmed.

   and whatever else it wants to say.

The frontend can skip step 2 and show the user something else if they don't want to support resends.

The confirmation email is not sent and the resend token is not set if a PUT does not update the address and the email has no earlier error.  In those cases, the resulting email will *still* have no error, so there's nothing to confirm.  In the updated-address case, the updated email will have an 'unconfirmed' error, and in the earlier-error case, the updated email will still have that earlier error; the confirmation email gives you a way to clear both.

Also:

* Add `email.id` (to support `/email/{id}`).
* Require `email.person` and `updateEmailUserPassword.person` (to avoid orphaned addresses).
* Move `/email/{id}/confirm` to `/emails/confirm`.  The backend will have to lookup the ID using the token.  The `{id}` version works well enough for PUT, but doesn't work for POST, because the caller has to   setup base-url before they know the email ID.  The workflow would have been:

    1. User POSTs a new email.
    2. User gets a confirmation email wit a confirmation URL.
    3. User follows confirmation URL.
    4. Code behind confirmation URL extracts the token from the URL, determines the email ID somehow, and POSTs the token to `/email/{id}/confirm`.

  but "code behind confirmation URL determines the email ID somehow" is unclear, while "API backend determines the email ID from the token" is possible (because the backend controls the token generation, could use a backing database, etc.).

A fixup for #147.

### DIFF
--- a/public.json
+++ b/public.json
@@ -4020,8 +4020,71 @@
           "200": {
             "description": "Email response.  An email with a confirmation URL will be mailed to the address, and new emails will be unconfirmed until the token is submitted via emailConfirmation.",
             "schema": {
+              "$ref": "#/definitions/updatedEmail"
+            }
+          },
+          "default": {
+            "description": "Unexpected error.",
+            "schema": {
+              "$ref": "#/definitions/errorModel"
+            }
+          }
+        }
+      }
+    },
+    "/emails/confirm": {
+      "post": {
+        "summary": "Complete an in-progress email confirmation.",
+        "operationId": "emailConfirmation",
+        "tags": [
+          "email"
+        ],
+        "parameters": [
+          {
+            "name": "token",
+            "in": "body",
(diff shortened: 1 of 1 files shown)


davidmcatee-azure:
This all makes sense.

PR #152, opened 2016-08-27
public.json: CRUD for heros

wking:
Give marketing a way to maintain a dozen or so hero promotions for a homepage slide show.


wking:
The homepage is now managed in WordPress, so there's no longer a need for this.

PR #154, opened 2016-09-08
add storage-climate to the product definition

MercenaryJosh:
see azurestandard/beehive#2179

### DIFF
--- a/public.json
+++ b/public.json
@@ -6125,6 +6125,16 @@
         "country-of-origin": {
           "description": "country of origin for this product",
           "type": "string"
+        },
+        "storage-climate": {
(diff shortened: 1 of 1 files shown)


public.json
```diff
@@ -6125,6 +6125,16 @@
...
           "type": "string"
+        },
+        "storage-climate": {
+          "description": "the climate at which this product must be stored",
+          "type": "string",
+          "enum": [
+            "Dry",
+            "Chilled",
+            "Frozen",
+            "Greenhouse"
```
wking: The rest of the API uses lowercase slugs for enums, so I think we want “Dry” → “dry”, etc.  Other than that, this PR looks good.

PR #155, opened 2016-09-08
Add packs and unit to packagedProduct definition

colin-merc-tech:
ref https://github.com/azurestandard/beehive/issues/2133

### DIFF
--- a/public.json
+++ b/public.json
@@ -5939,6 +5939,14 @@
           "description": "size of the product (e.g. 12 oz., or 6x12 oz.)",
           "type": "string"
         },
+        "packs": {
+          "description": "the dimensions of package sizes (e.g. \"4\", \"4x32.5\", or \"2x4x32.5\")",
+          "type": "string"
+        },
+        "unit": {
+          "description": "the unit of measurement",
+          "type": "string"
+        },
         "weight": {
           "minimum": {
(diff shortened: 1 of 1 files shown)


wking:
This looks reasonable to me, but it makes `packagedProductAudit` largely superfluous.  Can we shift `audited` over as `packs-audited` or some such and only return `packs-audited` to folks with a `change-packaged-product` permission?  Then we can drop `packagedProductAudit`, and roll the current audit endpoints into the usual `GET /packages` and a new `POST /packaged-product/{id}`.  The POST could 500 folks trying to change anything besides unit/packs/name, or we could just ignore the other fields for now.

tompetersjr:
@colin-merc-tech:  What are your thoughts on wking's comment.  I'm not sure I follow or understand the impact time wise to make such a change.

public.json
```diff
@@ -6011,6 +6019,8 @@
       "required": [
         "code",
         "size",
+        "packs",
+        "unit",
```
wking: Are we sure we want to require these in all cases and not just if the packaging has been audited?  For example, we [only set `unit` in the price for audited packaging][1].

[1]: https://github.com/azurestandard/beehive/blob/f79b8af4a60065e958ffc5a1f853217b225c2141/apps/api/views/price.py#L39

PR #156, opened 2016-09-08
public.json: Per-storage entries for order.cases-shipped

wking:
Drop coordinators are used to seeing these counts broken down by climate, with [separate counts for greenhouse, frozen, and dry/chilled][1].

Also unifies the storage types so `drop.storage` and `product.storage` (was `product.storage-climate`) share the same enum.  Before we land this, we should check and see if drops can opt in/out of storing greenhouse products.

[1]: https://github.com/azurestandard/beehive/blob/226ec6956970fa83c220c4ee4b05d6e35cf27294/apps/shipping/templates/shipping/pdfs/cover_sheets.html#L83-L103

### DIFF
--- a/public.json
+++ b/public.json
@@ -6460,14 +6460,13 @@
           "description": "parcel-carrier delivery location (unset for truck orders)",
           "$ref": "#/definitions/address"
         },
+        "cases-shipped": {
+          "description": "Quantity of shipped boxes and bulk stock",
+          "$ref": "#/definitions/orderCases"
+        },
         "notes": {
           "description": "Warehouse packing instructions",
           "type": "string"
-        },
-        "cases-shipped": {
-          "description": "quantity of shipped boxes and bulk stock",
-          "type": "integer",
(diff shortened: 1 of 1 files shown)


wking:
In today's meeting, David Stelzer said that drop coordinators cannot toggle `greenhouse` independently in their drop preferences (it's covered by `dry`).  We could restrict `drop.storage` to reflect this, but the phrasing was awkward.  The website currently has [special][1] [handling][2] for the configurable choices, so we can set `greenhouse` based on `dry` without requiring special casing in the backend.  And we can make the backend 400 drop pushes where `storage` contained one of `dry` or `greenhouse` but not the other.

[1]: https://github.com/azurestandard/website/blob/v1.0.9/app/views/partials/drop_preferences_form.html#L140-L157
[2]: https://github.com/azurestandard/website/blob/v1.0.9/app/scripts/directives/as-drop-preferences.js#L42

PR #158, opened 2016-09-13
Registration confirmation also performs login.

davidmcatee-azure:
This change would allow clients to treat a successful /registration/confirm response as a successful login of the user that was activated. Login would only occur if activation was successful, so it would only work once per token as activation does.

This would allow for a more streamlined registration flow, as the front end will be able to immediately engage the user with links to add a drop, fill out profile information, etc., without them having to sign in first.

If possible, I would like to get this in as a precursor to azurestandard/website#653.

### DIFF
--- a/public.json
+++ b/public.json
@@ -4528,7 +4528,7 @@
             "required": true,
             "type": "object",
             "schema": {
-              "$ref": "#/definitions/registration"
+              "$ref": "#/definitions/registrationRequest"
             }
           }
         ],
@@ -4565,9 +4565,9 @@
         ],
         "responses": {
           "200": {
-            "description": "session response",
+            "description": "registration confirmation response",
             "schema": {
-              "$ref": "#/definitions/session"
+              "$ref": "#/definitions/registrationConfirmResponse"
             }
           },
           "default": {
@@ -6971,7 +6971,7 @@
         "payment-method"
       ]
     },
-    "registration": {
+    "registrationRequest": {
       "description": "A request for registering a new person",
       "type": "object",
       "properties": {
@@ -7017,6 +7017,18 @@
         "person",
         "email"
       ]
+    },
+    "registrationConfirmResponse": {
+      "description": "The response to a successful registration confirmation, including the person who registered and the session information",
+      "type": "object",
+      "properties": {
+        "person": {
+          "$ref": "#/definitions/person"
+        },
+        "session": {
+          "$ref": "#/definitions/session"
+        }
+      }
     },
     "resendRegistrationEmail": {
       "description": "A request for another registration email",
--- a/public.json
+++ b/public.json
@@ -7027,7 +7027,11 @@
         },
         "session": {
           "$ref": "#/definitions/session"
-        }
+        },
+        "required": [
+          "person",
+          "session"
+        ]
       }
     },
     "resendRegistrationEmail": {
--- a/public.json
+++ b/public.json
@@ -4548,7 +4548,8 @@
     },
     "/registration/confirm": {
       "post": {
-        "summary": "Complete an in-progress registration",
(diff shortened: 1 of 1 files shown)


public.json
```diff
@@ -4564,9 +4565,9 @@
         ],
         "responses": {
           "200": {
-            "description": "person response",
+            "description": "session response",
             "schema": {
-              "$ref": "#/definitions/person"
+              "$ref": "#/definitions/session"
```
wking: Not all registration clients may want to use session-based auth, and they may want to know who they just confirmed ;).  Can we rename `definitions/registration` to `definitions/registrationRequest` and add a new `registrationResponse` type with `person` and `session` properties?

Alternatively, we can keep the response type unchanged here, and just document that we're also logging you in.  Clients using session-based auth can then hit `/session` if they want the session information.

davidmcatee-azure: > Can we rename definitions/registration to definitions/registrationRequest and add a new registrationResponse type with person and session properties?

I almost did it that way but went simpler instead. I figured if you're getting a session cookie then you might as well get a session response. Since the response contains the person id you could always ignore the fact that you're getting a session response/cookie and use `/person` to get the full person if needed.  Having `registrationResponse` with `person` and `session` properties would be more explicit and probably more understandable, so I'm OK with that if you're uncomfortable with the simple session response.

wking: On Tue, Sep 13, 2016 at 03:56:23PM -0700, David McAtee wrote:
> >              "schema": {
> > -              "$ref": "#/definitions/person"
> > +              "$ref": "#/definitions/session"
>
> > Can we rename definitions/registration to
> > definitions/registrationRequest and add a new registrationResponse
> > type with person and session properties?
>
> I almost did it that way but went simpler instead. I figured if
> you're getting a session cookie then you might as well get a session
> response.

I'm fine returning the session response (or not).  I just want to
return the person for clients that don't use cookies.

> Since the response contains the person id you could always ignore
> the fact that you're getting a session response/cookie and use
> `/person` to get the full person if needed.

But you'd need to authenticate that /person request.  If you're not
listening for and storing cookies in the response headers, you may not
be able to authenticate that follow-up request.

> Having `registrationResponse` with `person` and `session` properties
> would be more explicit and probably more understandable, so I'm OK
> with that if you're uncomfortable with the simple session response.

This makes us both happy ;).

davidmcatee-azure: > But you'd need to authenticate that /person request.  If you're not
> listening for and storing cookies in the response headers, you may not
> be able to authenticate that follow-up request.

I hadn't thought of that.  I'll make the changes.

`registrationResponse` sounds like it's the response corresponding to the `registrationRequest`. How about leaving `registration` as is and adding `registrationConfirmResponse`?

wking: On Tue, Sep 13, 2016 at 04:17:57PM -0700, David McAtee wrote:
> `registrationResponse` sounds like it's the response corresponding
> to the `registrationRequest`. How about leaving `registration` as is
> and adding `registrationConfirmResponse`?

I think ‘registration’ is too vague if we have multiple registration
structures.  Can we use registrationRequest and
registrationConfirmResponse?

davidmcatee-azure: Sounds good.

davidmcatee-azure:
Let me know if this looks good now.

public.json
```diff
@@ -7017,6 +7018,18 @@
...
+    "registrationConfirmResponse": {
+      "description": "The response to a successful registration confirmation, including the person who registered and the session information",
+      "type": "object",
+      "properties": {
+        "person": {
+          "$ref": "#/definitions/person"
+        },
+        "session": {
+          "$ref": "#/definitions/session"
+        }
```
wking: Require both of these?

PR #159, opened 2016-09-21
Revert 155 mt/add unit packs

tompetersjr:
This is being opened as I merged the original by mistake.  That was reverted.  Please see original conversation.

https://github.com/azurestandard/api-spec/pull/155


wking:
Replaced by #160.

PR #160, opened 2016-09-21
Add packs and unit to packagedProduct definition

wking:
Reopening #155, since GitHub closed that one after an accidental merge (which we've since rolled back).


wking:
I've rebased onto master, [dropped the `required` entries][1], and tweaked the descriptions, and added some motivation to the commit message in 4fba092 → 4900af9.

[1]: https://github.com/azurestandard/api-spec/pull/155#discussion_r83546087

PR #167, opened 2016-10-26
Add definitions and endpoints for BarTender integration.

caseybessette:
These will be used for managing BarTender print configurations in the
internal website.  There is a special endpoint for testing a print
configuration.  Also, minimal definitions of warehouses and printers
(PrinterOptions) are exposed in order to create print configurations.

### DIFF
--- a/public.json
+++ b/public.json
@@ -45,6 +45,588 @@
     "application/json"
   ],
   "paths": {
+    "/bartender-hosts": {
+      "get": {
+        "summary": "Returns all BarTender hosts from the system that the user has access to",
+        "operationId": "findBartenderHosts",
+        "tags": [
+          "bartender"
+        ],
+        "parameters": [
+          {
+            "name": "limit",
+            "in": "query",
+            "description": "maximum number of results to return",
+            "required": false,
+            "type": "integer",
+            "format": "int32"
+          },
+          {
+            "name": "start",
+            "in": "query",
+            "description": "offset to the first result to return.  Use negative numbers to offset from the end of the result list.",
+            "required": false,
+            "type": "integer",
+            "format": "int32"
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "bartenderHost response",
+            "schema": {
+              "type": "array",
+              "items": {
+                "$ref": "#/definitions/bartenderHost"
+              }
+            },
+            "headers": {
+              "Count": {
+                "description": "total number of matching results (how many you'd get if you could set an infinite `limit`)",
+                "type": "integer",
+                "format": "int32",
+                "minimum": 0
+              }
+            }
+          },
+          "default": {
+            "description": "unexpected error",
+            "schema": {
+              "$ref": "#/definitions/errorModel"
+            }
+          }
+        }
+      },
+      "post": {
+        "summary": "Create a new BarTender host",
+        "operationId": "addBartenderHost",
+        "tags": [
+          "bartender"
+        ],
+        "parameters": [
+          {
+            "name": "host",
+            "in": "body",
+            "description": "BarTender host to add",
+            "required": true,
+            "schema": {
+              "$ref": "#/definitions/bartenderHost"
+            }
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "BarTender host response",
+            "schema": {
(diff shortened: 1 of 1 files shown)


public.json
```diff
@@ -45,6 +45,588 @@
...
+            "schema": {
+              "$ref": "#/definitions/errorModel"
+            }
+          }
+        }
+      }
+    },
+    "/bartender-print-configuration/{id}/actions/test": {
+      "post": {
+        "summary": "Create a test BarTender print request using this print configuration",
```
wking: Can we make this `print` instead of `test`?  It seems like this endpoint results in a physical label, but no changes to the persistent Beehive tables.

public.json
```diff
@@ -5151,9 +5733,241 @@
...
+              "$ref": "#/definitions/errorModel"
+            }
+          }
+        }
+      }
     }
   },
   "definitions": {
+    "bartenderHost": {
+      "description": "a BarTender host/instance",
```
wking: It's slightly odd to have `BarTender` in the description and `bartender` in `bartenderHost`.  For the camelCase tags can we go with `barTender*`?

public.json
```diff
@@ -5151,9 +5733,241 @@
...
+    "bartenderTemplate": {
+      "description": "a BarTender template/document",
+      "type": "object",
+      "properties": {
+        "id": {
+          "type": "integer",
+          "format": "int64"
+        },
+        "path": {
+          "type": "string"
```
wking: Can we add a description here?  Something like:

> Path to the BarTender template on the BarTender host filesystem.

public.json
```diff
@@ -5151,9 +5733,241 @@
...
+        "host": {
+          "description": "the BarTender host's code",
+          "type": "string"
+        },
+        "printer": {
+          "type": "integer",
+          "format": "int64"
+        },
+        "template": {
+          "type": "integer",
```
wking: This needs a description like:

> The BarTender template ID.

public.json
```diff
@@ -5151,9 +5733,241 @@
...
+        "id": {
+          "type": "integer",
+          "format": "int64"
+        },
+        "host": {
+          "description": "the BarTender host's code",
+          "type": "string"
+        },
+        "printer": {
+          "type": "integer",
```
wking: This could use a description like:

> The printer ID.  The referenced printer must support the `bartender` format and delivery methods.

public.json
caseybessette: How about `test-print`?  I think it would be nice to have the endpoint clearly designate that this is for testing and not for sending regular print jobs.

public.json
caseybessette: Will do.

public.json
caseybessette: This is slightly confusing.  "BarTender" is a trade-marked name, and it is one word.  So the Beehive model is `BarTenderHost` but the database table is `bartender_host`.  The description has the English name "BarTender host" but the camelCase tags only capitalize for distinct words (like the database table), so `bartenderHost`.  I can change it, but that was my reasoning. :)

public.json
caseybessette: Will do.

public.json
caseybessette: Will do.

caseybessette:
> I'm also missing the interface for `Piece.bartender_print_configurations`.  Is that getting setup somewhere else?  In a later API PR?  If we don't plan on adding an API for that association, it's probably worth documenting the missing link somewhere in this PR (the configuration model description?).

The interface for `Piece.bartender_print_configurations` is going to be on the piece page in Beehive and will not be configured via the API.  I will add a note to the configuration model description.

public.json
```diff
@@ -5151,9 +5733,244 @@
...
+          "type": "string"
+        }
+      },
+      "required": [
+        "id",
+        "path"
+      ]
+    },
+    "bartenderPrintConfiguration": {
+      "description": "A BarTender print configuration for BarTender print requests.  These configurations are added to warehouse pieces (which are related to individual stock) in Beehive and are used to route print requests and to tell BarTender which template should be used.  A print request will be created for each print configuration printer associated with a stock location's warehouse.",
```
wking: “for each print configuration printer” → “for each print configuration”.

PR #169, opened 2016-10-28
public.json: Move /emails/resend to /email/{id}/resend

### LINKED ISSUE #168: move resend to /emails/{id}/resend
@wking

Related to multiple email support ([website issue 624](https://github.com/azurestandard/website/issues/624)):

**Resending the confirmation email:**

With the current api spec, there seem to be two ways to achieve this:

1. POST to `/emails/resend` - which requires a token that is only returned after initially creating the email (and is thus not possible to get again after page refresh, unless we stored it in localstorage or something, which at that point seems to overly complicate the task).
2. PUT to `/email/{id}` - which requires the user's password (arguably bad UX for such a simple task as having a confirmation email resent).

Since a POST request to `/emails/resend` seems specifically intended for this task it appears the recommended approach. However, a PUT request to `/email/{id}` is much easier since it does not require a token. Perhaps there may be a way to drop the token requirement for `/emails/resend` or otherwise PUT to ``/email/{id}` without a password? Anything I'm missing here?

Also, I don't see why a PUT to `/email/{id}/resend` with the email's ID wouldn't work (since the ID is returned on POST (creation) of the email via `/emails/create` and on the loading of emails (via GET to `/emails`), so it's accessible in all cases the front-end would need it for making a PUT to `/emails/{id}/resend`). Reading over the initial implementation of the API it seems this was the approach... I've read the [PR notes here](https://github.com/azurestandard/api-spec/pull/149#issue-172022037) but since there's access to the ID I'm not sure on why this approach wouldn't work.

wking:
Also remove the empty-security override, make resend tokens optional, and typo-fix "resent" -> "re-sent".

The old ID-less version matched /emails/confirm.  Confirmation shouldn't rely on the email ID for the reasons described in #149.  However, the frontend triggering resends should be able to use both tokens and the usual basic/cookie credentials to trigger a resend.  By shifting /emails/resend to /email/{id}/resent, we give frontend a way to pass the ID without passing the token (which encodes the ID, among other things).  Passing the token still works; folks who pass the token will now need to track both the token and the email ID, but both are part of the updatedEmail payload so that shouldn't be a problem.

Fixes #168.

### DIFF
--- a/public.json
+++ b/public.json
@@ -4647,38 +4647,6 @@
         }
       }
     },
-    "/emails/resend": {
-      "post": {
-        "summary": "Resend the confirmation email to the user.",
-        "operationId": "resendEmailConfirmation",
-        "tags": [
-          "email"
-        ],
-        "security": [],
-        "parameters": [
-          {
-            "name": "resend",
-            "in": "body",
-            "required": true,
-            "schema": {
-              "$ref": "#/definitions/resendRegistrationEmail"
-            }
-          }
-        ],
-        "responses": {
-          "204": {
-            "description": "Confirmation email resent",
-            "schema": {}
-          },
-          "default": {
-            "description": "Unexpected error.",
-            "schema": {
-              "$ref": "#/definitions/errorModel"
-            }
-          }
-        }
-      }
-    },
     "/email/{id}": {
       "get": {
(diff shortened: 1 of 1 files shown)


igregson:
Looks good.

PR #170, opened 2016-10-31
Add printer filters.

caseybessette:


public.json
```diff
@@ -5750,6 +5750,38 @@
...
+            "description": "printer delivery method",
+            "required": false,
+            "type": "array",
+            "items": {
+              "type": "string",
+              "enum": [
+                "http",
+                "cups",
+                "bartender"
+              ]
```
wking: We should pull this enum out into it's own field (`printerDeliveryType`?) so we can use it here and in the `printer` definition.  For an existing example of this approach, see `orderFeeType`.  This also applies to the format enum you're adding a filter for below.

wking:
Implementation in azurestandard/beehive#1748.

PR #176, opened 2016-11-16
public.json: Use country in address and address in ip2location

wking:
Reduce the number of places where we define similar-but-not-identical models for the same thing.  Use `?inline=…` to allow callers to push joins to the backend, avoiding multiple API calls for client-side joins.

CC @MercenaryJosh, @tompetersjr, since this touches the API which just landed in #175.

CC @davidmcatee-azure, since I'm not sure how much trouble updating the current frontend's address consumers to follow the address.country-name → address.country change will be.

### DIFF
--- a/public.json
+++ b/public.json
@@ -4235,6 +4235,17 @@
             "collectionFormat": "csv"
           },
           {
+            "name": "inline",
+            "in": "query",
+            "description": "for convenience, include a full representation of the inlined property.  Currently supports \"country\".",
+            "required": false,
+            "type": "array",
+            "items": {
+              "type": "string"
+            },
+            "collectionFormat": "csv"
+          },
+          {
             "name": "limit",
             "in": "query",
             "description": "maximum number of results to return",
@@ -5950,6 +5961,17 @@
             "description": "ipv4 address to lookup",
             "required": true,
             "type": "string"
+          },
+          {
+            "name": "inline",
+            "in": "query",
+            "description": "for convenience, include a full representation of the inlined property.  Currently supports \"address.country\".",
+            "required": false,
+            "type": "array",
+            "items": {
+              "type": "string"
+            },
+            "collectionFormat": "csv"
           }
         ],
         "responses": {
@@ -8243,7 +8265,8 @@
         "postal-code": {
           "type": "string"
         },
-        "country-name": {
+        "country": {
+          "description": "ISO 3166-1 alpha-3 code for the country.  On push, you can set the alpha-2 code or case-insensitve name instead, or you can use an inline country object.",
           "type": "string"
         },
         "preference": {
@@ -8254,7 +8277,7 @@
       },
       "required": [
         "locality",
-        "country-name"
+        "country"
       ]
     },
     "country": {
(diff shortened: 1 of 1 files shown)


MercenaryJosh:
The ip2location endpoint is only going to be used for determining a users location by their ip address, so I don't think the inline option is necessary. This endpoint will also never return most of the address data, so it would be misleading to suggest that it may.

davidmcatee-azure:
> how much trouble updating the current frontend's address consumers to follow the address.country-name → address.country change

It would be trivial on GET using the `inline` parameter, but will the POST to `/addresses` be affected?  It looks like the api-spec is already inaccurate there because it says it requires a proper `address`, but we also include phone, text, and preference.

wking:
On Wed, Nov 16, 2016 at 10:26:11AM -0800, Joshua Hefner wrote:
> The ip2location endpoint is only going to be used for determining a
> users location by their ip address, so I don't think the inline
> option is necessary.

You get the *location* from geo.  You'd only inline the country if you
wanted to display the country name (maybe never?).

> This endpoint will also never return most of the address data, so it
> would be misleading to suggest that it may.

Mapping ip2location fields across this commit:

* country-short → address.country.iso3 ?
* country-long → address.country.name
* region → address.region
* city → address.locality
* zipcode → address.postal-code

Address fields which ip2location will not fill: name, post-office-box,
extended-address, street-address, preference.

Filling about half of the address fields is enough for me to not want
to create a new model, and certainly not a new model that uses
different names for existing properties (e.g. city vs. locality).

wking:
On Wed, Nov 16, 2016 at 10:36:45AM -0800, David McAtee wrote:
> It would be trivial on GET using the `inline` parameter, but will
> the POST to `/addresses` be affected?

You'd need to move country-name → name, but I've tried to cover the
existing “use the case-insensitive name, iso-3, or iso-2 as you wish”
functionality.

> It looks like the api-spec is already inaccurate there because it
> says it requires a proper `address`, but we also include phone,
> text, and preference.

We already cover preference [1].  I agree that the phone types seem to
be missing; I'll file a separate PR for those.

[1]: https://github.com/azurestandard/api-spec/blob/f9418cc31f489021bc3ab9cfcf59340bef9ce34e/public.json#L8249-L8253

davidmcatee-azure:
OK, i didn't see the preference there.  So a previous POST with
```json
{
  country-name: "UNITED STATES",
  ...
}
```
would just become:
```json
{
  country: {
    name: "UNITED STATES"
  },
  ...
}
```

Is that correct?

wking:
On Wed, Nov 16, 2016 at 10:47:03AM -0800, David McAtee wrote:
> So a previous POST with
> ```json
> {
>   country-name: "UNITED STATES",
>   ...
> }
> ```
> would just become:
> ```json
> {
>   country: {
>     name: "UNITED STATES"
>   },
>   ...
> }
> ```
> 
> Is that correct?

Yup, and the way it's written [1], you could also do:

  {
    "country": "UNITED STATES",
    …
  }

or:

  {
    "country": "US",
    …
  }

or:

  {
    "country": "USA",
    …
  }

[1]: https://github.com/azurestandard/api-spec/pull/176/files#diff-380fafdca01526afaf5cbc26fb2f0983R8269
     “On push…”

davidmcatee-azure:
Thanks for clarifying. I didn't see that either.. I find it really hard to read the api-spec.

The frontend changes would be easy then.

tompetersjr:
If this is not a significant effort I am fine with it.  I'm re-assigning to you David as this will need to be coordinated with Frontend / Backend releases.

PR #177, opened 2016-11-16
remove ipaddress from /ip2location

MercenaryJosh:
The ip2location endpoint must detect the user's ip address as well as
return the location data.

### DIFF
--- a/public.json
+++ b/public.json
@@ -5939,19 +5939,10 @@
         }
       }
     },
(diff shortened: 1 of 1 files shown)


wking:
On Wed, Nov 16, 2016 at 03:12:50PM -0800, Joshua Hefner wrote:
> The ip2location endpoint must detect the user's ip address as well
> as return the location data.

This looks good to me.

PR #183, opened 2016-12-09
public.json: Change registration base-url to authentication-base-url

wking:
And `confirmation-base-url`, since we now have two different tokens that we might be mailing the registrant.

I've gone with “channel” instead of “email” in the description, because we might to authenticate additional channels besides email (e.g. SMS) in the future.

### DIFF
--- a/public.json
+++ b/public.json
@@ -7891,11 +7891,6 @@
       "description": "A request for registering a new person",
       "type": "object",
       "properties": {
-        "base-url": {
-          "description": "Redirect URL for the confirmation UI.  The confirmation email points customers at {base-url}{confirmation-token} to confirm their registration.",
-          "type": "string",
-          "format": "url"
-        },
         "person": {
           "description": "The registrant's personal information.  'name' is required.",
           "type": "object",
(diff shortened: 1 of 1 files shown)


public.json
```diff
@@ -7930,12 +7925,35 @@
...
+            "type": "string",
+            "enum": [
+              "registration",
+              "checkout",
+              "catalog-request"
+            ]
+          }
+        },
+        "authentication-base-url": {
+          "description": "Redirect URL for the authentication UI.  The authentication notification points customers at {base-url}{authentication-token} to authenticate a new session.",
```
davidmcatee-azure: `{authentication-base-url}{authentication-token}`

public.json
```diff
@@ -7930,12 +7925,35 @@
...
+          "type": "string",
+          "format": "url"
+        },
+        "confirmation-base-url": {
+          "description": "Redirect URL for the channel-confirmation UI.  The channel-confirmation notification points customers at {base-url}{confirmation-token} to confirm the notification channel.",
```
davidmcatee-azure: `{confirmation-base-url}{confirmation-token}`

wking:
On Fri, Dec 09, 2016 at 03:53:55PM -0800, David McAtee wrote:
> `{authentication-base-url}{authentication-token}`
> …
> `{confirmation-base-url}{confirmation-token}`

Fixed with 0c7235c → 33c0652.

PR #184, opened 2016-12-13
Allow sorting drops by distance.

### LINKED ISSUE #179: Sort /drops by proximity to geo coordinates
Update the /drops endpoint to accept geo coordinates as a parameter. If they are provided, sort the results by proximity to the given coordinates.

caseybessette:
The general `sort` query parameter means additionally sort fields can be
added in the future.  Since the distance needs to be calculated from a
lat + lon input, a special format is used here.

Fixes #179.

### DIFF
--- a/public.json
+++ b/public.json
@@ -678,6 +678,11 @@
             "collectionFormat": "csv"
           },
           {
+            "name": "sort",
+            "in": "query",
+            "description": "Order results by this field.  Only one special sort field is supported, \"distance(lat:45.45|lon:-121.13)\", which will sort drops by increasing distance from the lat and lon combination."
+          },
+          {
             "name": "limit",
             "in": "query",
             "description": "maximum number of results to return",


public.json
```diff
@@ -692,6 +692,11 @@
             "required": false,
             "type": "integer",
             "format": "int32"
+          },
+          {
+            "name": "sort",
+            "in": "query",
+            "description": "Order results by this field.  Only one special sort field is supported, \"distance(lat:123|lon:-321)\", which will sort drops by their distance from the lat and lon combination."
```
caseybessette: Will do.

caseybessette:
@wking Updated lat,lon example and commit message.

wking:
I just pushed ccda712 → d3c6255, moving the `sort` option [above the pagination options][1] and mentioning that the sort is by increasing distance.  Does that look ok?

[1]: https://github.com/azurestandard/api-spec/pull/184#discussion_r92265824

PR #185, opened 2016-12-15
Add order parcel carrier shipping endpoint and definition

colin-merc-tech:
Opening this PR to give us a place to discuss the format for the new endpoint we're going to create to handle retrieving parcel carrier shipping estimates for orders. 

The commit I've included will end up looking something this

```
[
    {
        "order": 13414,
        "service-name": "UPS Next Day Air",
        "shipping-days": 1,
        "cost": 55.00,
        "guaranteed": false,
    },
    {
        "order": 13414,
        "service-name": "UPS 2-Day Express",
        "shipping-days": 2,
        "cost": 25.00,
        "guaranteed": true,
    },
]
```

### DIFF
--- a/public.json
+++ b/public.json
@@ -7297,8 +7297,17 @@
         "estimated-time"
       ]
     },
+    "shippingClimate": {
+      "description": "An environment for shipping products.  'all' is a shipping climate for products of all types.  'dry' includes both 'dry' and 'greenhouse' storage environments.  'chilled/frozen' includes both 'chilled' and 'frozen' storage environments.",
+      "type": "string",
+      "enum": [
+        "all",
+        "dry",
+        "chilled/frozen"
+      ]
+    },
     "storage": {
-      "description": "An environment for storing products",
+      "description": "An environment for storing products.  The related to shippingClimate is less detailed, because stock is generally in transit for less time than it is in storage.",
       "type": "string",
       "enum": [
         "dry",
@@ -8116,6 +8125,10 @@
           "description": "Dollar value for the fee",
           "type": "number",
           "format": "float"
+        },
+        "climate": {
+          "description": "The shipping climate for order lines that the fee applies to.",
+          "$ref": "#/definitions/shippingClimate"
         },
         "notes": {
           "description": "Additional information about the reason or amount for the fee",
--- a/public.json
+++ b/public.json
@@ -4249,6 +4249,43 @@
         }
       }
     },
+    "/order/{id}/parcel-carrier-fee-estimates": {
+      "get": {
+        "summary": "Returns all parcel carrier shipping options and their costs for an order ID",
+        "operationId": "listOrderParcelCarrier",
+        "tags": [
+          "order",
+          "parcel-carrier"
+        ],
+        "parameters": [
+          {
+            "name": "id",
+            "in": "path",
+            "description": "ID of order to retrieve parcel carrier options for.",
+            "required": true,
+            "type": "integer",
+            "format": "int64"
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "Order parcel carrier estimate response.",
+            "schema": {
+              "type": "array",
+              "items": {
+                "$ref": "#/definitions/orderParcelCarrierEstimate"
+              }
+            }
+          },
+          "default": {
+            "description": "Unexpected error.",
+            "schema": {
+              "$ref": "#/definitions/errorModel"
+            }
+          }
+        }
+      }
+    },
     "/vendors": {
       "get": {
         "summary": "Returns all vendors from the system that the user has access to",
@@ -7297,6 +7334,16 @@
         "estimated-time"
       ]
     },
+    "shippingService": {
+      "description": "A parcel-carrier service.  The listing below is ordered by carrier and then by increasing duration.",
+      "type": "string",
+      "enum": [
+        "UPS 2nd Day Air",
+        "UPS Ground",
+        "USPS Priority Mail Express",
+        "USPS Priority Mail"
+      ]
+    },
     "shippingClimate": {
       "description": "An environment for shipping products.  'all' is a shipping climate for products of all types.  'dry' includes both 'dry' and 'greenhouse' storage environments.  'chilled/frozen' includes both 'chilled' and 'frozen' storage environments.",
       "type": "string",
@@ -7883,6 +7930,10 @@
           "description": "parcel-carrier delivery location (unset for truck orders)",
           "$ref": "#/definitions/address"
         },
+        "parcel-carrier-services": {
+          "description": "Parcel-carrier services selected for the order.",
+          "$ref": "#/definitions/orderParcelCarrierServices"
+        },
         "cases-shipped": {
           "description": "Quantity of shipped boxes and bulk stock",
           "$ref": "#/definitions/orderCases"
@@ -7934,6 +7985,10 @@
         "address": {
           "description": "parcel-carrier delivery location (unset for truck orders)",
           "$ref": "#/definitions/address"
+        },
+        "parcel-carrier-services": {
+          "description": "Parcel-carrier services selected for the order.",
+          "$ref": "#/definitions/orderParcelCarrierServices"
         },
         "notes": {
           "description": "Additional information about the order, including special instructions for Azure's pickers",
@@ -8130,6 +8185,10 @@
           "description": "The shipping climate for order lines that the fee applies to.",
           "$ref": "#/definitions/shippingClimate"
         },
+        "shipping-service": {
+          "description": "The shipping service name.",
+          "$ref": "#/definitions/shippingService"
+        },
         "notes": {
           "description": "Additional information about the reason or amount for the fee",
           "type": "string"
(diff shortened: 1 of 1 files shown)


public.json
```diff
@@ -7623,6 +7668,41 @@
         "small-order"
       ]
     },
+    "orderParcelCarrierService": {
+      "description": "A Parcel carrier shipping service for a given order.",
+      "type": "object",
+      "properties": {
```
davidmcatee-azure: Should we have an integer identifier as well? We'll need a way to pick one of these when we PUT an order.

colin-merc-tech: The service types won't be persisted in any form in our database, so we'll need to rely on a service-name to service info lookup to handle the selection. So passing along the service-name should be fine, as it'll be unique to the service it's self.

davidmcatee-azure:
Can we flesh out how the /order/{id} operations would look?  If I want to choose one of the parcel carrier service options, how would I PUT the order?  By adding the parcel carrier (with only the service name required) as a property to the order I'm putting: `{ "parcel-carrier": { "service-name" : "UPS Next Day Air" } }`?

And when I GET an order, I would expect the parcel carrier information to be included, except for the cost (which will be an orderFee returned from /order-fees). Is that possible? Ideally we shouldn't query any parcel carrier APIs in an order GET.

colin-merc-tech:
>Can we flesh out how the /order/{id} operations would look?

Definitely. I think we'll need Casey to jump in here as well. Apparently he's working on breaking order fees out into it's own API, and I'm guessing that's where the majority of this logic will land. 

@caseybessette will order fee information still exist within the order API (/order/{id}) after your changes?

colin-merc-tech:
>Ideally we shouldn't query any parcel carrier APIs in an order GET.
This is an important goal for this. We want to keep estimate calls to a minimum.

caseybessette:
@colin-merc-tech You can see my changes here: https://github.com/azurestandard/beehive/pull/2401  The order fees will not be provided by the order endpoint any longer.  The order fees *are* calculated at the first GET for the fees for an order.  They're cached after that until any changes are made to the UPS shipment.

colin-merc-tech:
> And when I GET an order, I would expect the parcel carrier information to be included, except for the cost (which will be an orderFee returned from /order-fees)

@davidmcatee-azure Can you explain when you'll need the carrier service name but not need the order fees? My first instinct is to bundle carrier service name with the order fee api. Not that we can't have in both places, but I feel like I'm missing something if that's a requirement.

davidmcatee-azure:
Sure. Currently when you choose Parcel Carrier as your shipment method, the cutoff timer looks like this:

screen shot 2016-12-20 at 10 01 45 am

Instead of "Parcel Carrier" we would like to say the service name, like "UPS Next Day Air". For that display we are not fetching the order fees yet. So at the very least I would like the order model to include something like:
```json
{
  "parcel-carrier": {
    "service-name": "UPS Next Day Air"
  }
}
```
The reason we split out the order fees in the first place was to delay the need to hit the third-party shipping APIs (originated from azurestandard/website#552).

If any more information about the chosen parcel carrier service is foreknown (not requiring an API call to the carrier), I figured we might as well return that with the order too.  But if nothing else is available, the service name would be sufficient. There are no plans to display `shipping-days` or `guaranteed` anywhere, aside from the initial shipment selection UI.  Having them available in case we decide to display that information after a shipment is selected would be nice, but not required at this stage.

caseybessette:
```
requests.get('http://a.com/api/order/1632106/parcel-carrier-fee-estimates', auth=auth).json()

  [{'climate': 'dry',
    'service': 'UPS Ground',
    'cost': 14.89},
   {'climate': 'dry',
    'service': 'UPS 2nd Day Air',
    #'guaranteed-days': '2',  <-- phase two?
    'cost': 27.05},
   ...
  ]

requests.get('http://a.com/api/order-fees?order=1632106', auth=auth)

[{'type': 'parcel-carrier', 'climate': 'dry', 'service': 'UPS Ground', 'order': 1632106, 'amount': 10.0, 'id': 896813},
 {'type': 'parcel-carrier', 'climate': 'frozen', 'service': 'UPS 2nd Day Air', 'order': 1632106, 'amount': 10.0, 'id': 896813}]

r = requests.get('http://a.com/api/order/1632106', auth=auth)

{
    'order': {
        'address': {...},
        'parcel-carrier-services': {
            'all': 'UPS Ground'
            #--or--
            'dry': 'UPS Ground',
            'frozen': 'UPS 2nd Day Air'
        }
    },
    'changes': [{
        'type': 'parcel-carrier-services-removed',
        'reason': 'order.address changed'
    }]
}

# Delete shipping order fees when any changes are made to order lines or order address.

r = requests.get('http://a.com/api/order-line/5932406', auth=auth)

{
    'order-line': {
        ...
    },
    'changes': [{
        'type': 'parcel-carrier-services-removed',
        'reason': 'order-line changed'
    }]
}

# Make the current warning on order-line into a "change"
```

public.json
```diff
@@ -7683,6 +7730,25 @@
         "notes": {
           "description": "Additional information about the reason or amount for the fee",
           "type": "string"
+        },
+        "climate": {
+          "description": "The storage climate of orderlines that the fee applies to.",
+          "enum": [
+            'all', 'dry', 'chilled/frozen'
```
wking: This enum appears in two places, and is similar to the existing `storage` definition.  Can we add a `shippingClimate` (or some such) definition near `storage` to cover these values?

I'm also not clear on `all`.  Can the enum definition get some docs about when we'd set that vs. a more specific shipping climate?  I expect it's when we're shipping everything (currently both `dry` and `chilled/frozen`) with the same frozen-compatible service.  But maybe folks expect `all` if the order only contains `dry` products?  Anyhow, it would be good to have this clarified.

public.json
```diff
@@ -3845,6 +3845,43 @@
         }
       }
     },
+    "/order/{id}/parcel-carrier-fee-estimates": {
+      "get": {
+        "summary": "Returns all parcel carrier shipping options and their costs for an order ID",
```
wking: “parcel carrier” → “parcel-carrier”.

public.json
```diff
@@ -3845,6 +3845,43 @@
...
+        "operationId": "listOrderParcelCarrier",
+        "tags": [
+          "order",
+          "parcel-carrier"
+        ],
+        "parameters": [
+          {
+            "name": "id",
+            "in": "path",
+            "description": "ID of order to retrieve parcel carrier options for",
```
wking: “parcel carrier” → “parcel-carrier”.

public.json
```diff
@@ -7444,6 +7481,11 @@
         "notes": {
           "description": "Warehouse packing instructions",
           "type": "string"
+        },
+        "parcel-carrier-services": {
+          "description": "parcel carrier services for the order.",
+          "type": "array",
+          "$ref": "#/definitions/orderParcelCarrierServices"
```
wking: I think you need `items` for an array schema:

```
"type": "array",
"items": {
  "$ref": "#/definitions/orderParcelCarrierServices"
}
```

public.json
```diff
@@ -7696,9 +7762,87 @@
...
+        "cost",
+        "climate"
       ]
     },
+    "orderParcelCarrierServices": {
+      "description": "Parcel Carrier Estimate to select a service for.",
+      "type": "object",
+      "properties": {
+        "all": {
+          "description": "The service to be used for all order lines.",
```
wking: Maybe `all` should be `default`?  For example, if we decide to split `chilled` and `frozen` in the future, you could set `default` and `frozen` entries and clearly have the default apply to both `dry` and `chilled`.  I don't think that's as obvious if someone were to set `all` and `frozen` (which one did they expect to apply to `frozen`?).

caseybessette:
To support delivery estimates, we will use this schema for the shipping estimates response:

```javascript
[
    {
        'service': 'UPS Ground',
        'cost': 21.01,
        'expires': '2017-06-23T13:00',
        'shipping-date': '2017-06-23',
        'earliest-delivery': '2017-06-26',
        'latest-delivery': '2017-06-28'
    },
    ...
]
```

wking:
On Thu, Jun 22, 2017 at 11:12:28AM -0700, Casey M. Bessette wrote:
>         'expires': '2017-06-23T13:00',

We don't currently call this out in the API spec, but all date-times
should have explicit timezones.  So this would be:

  'expires': '2017-06-23T13:00-07:00',

caseybessette:
This is the updated schema for the shipping estimates response:

```javascript
[
    {
        'service': 'UPS Ground',
        'cost': 21.01,
        'expires': '2017-06-23T13:00',
        'shipping-date': '2017-06-23',
        'earliest-delivery': '2017-06-26',
        'latest-delivery': '2017-06-28',
        'guaranteed': true,
        'requires-perishable-shipping-warning': true
    },
    ...
]
```

wking:
I've pushed 21d9186 → b4c7b64 which I think catches us up with the current implementation.  I've left [`guaranteed` and `requires-accepted-perishable-shipping-warning`][1] out for now, because they're still in flight with azurestandard/beehive#2950.

[1]: https://github.com/azurestandard/api-spec/pull/185#issuecomment-318697400

wking:
I'd confused myself about what the new estimates endpoint was returning.  I've just pushed b4c7b64 → a1f49da, which adjusts the estimates response from an array of an array of objects to just being an array of objects.

public.json
```diff
@@ -7297,8 +7334,29 @@
         "estimated-time"
       ]
     },
+    "shippingService": {
+      "description": "A parcel-carrier service.  The listing below is ordered by carrier and then by increasing duration.",
+      "type": "string",
+      "enum": [
+        "UPS Next Day Air",
```
caseybessette: We are not supporting `UPS Next Day Air` or `UPS Standard` at this point.

wking: > We are not supporting UPS Next Day Air or UPS Standard at this point.

I removed them with a1f49da → 17e5169.

PR #186, opened 2016-12-19
person: Add .internal-notes

wking:
Our backend has several ways to store notes associated with a given user:

1. In an append-only array of internal notes with no UI for editing existing notes (via the Django comments app).
2. In an internal, editable text field usually used for billing notes.
3. In an editable text field for default shipping instructions, which the customer and internal users can see.

This commit exposes (2).  Our immediate use-case is for recording promotion codes and similar at registration time so we can create credits for signing bonuses and the like.

Also two API fixups to match our current implementation, since some changes were made without a leading API-spec PR.

### DIFF
--- a/public.json
+++ b/public.json
@@ -7970,8 +7970,7 @@
         },
         "token": {
           "description": "Resend token.  Can be sent to resendRegistrationEmail to trigger a new registration notification.",
-          "type": "string",
-          "format": "url"
+          "type": "string"
         },
         "required": [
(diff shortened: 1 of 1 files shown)


tompetersjr:
This looks find to me.

tompetersjr:
@wking As long as this does not break any existing functionality (pretty sure it does not) let's go ahead and release when you have the supporting backend piece in place.

PR #188, opened 2017-01-15
Add product substitutions

shofetim:


wking:
I've taken a stab at rebasing this around the just-landed #107 with 17e9ecc → 649693b.  I've put `substitutions` in the “local” group (vs. the “… and descendants” group) because the implementation PR is deciding between [“leave this property unset” and “set this property to an array of IDs”][1].  For comparison, the `packaging` logic in the “…and descendents” is deciding between [“leave this property unset” and “set this property to an array of packagedProduct objects”][2].

[1]: https://github.com/azurestandard/beehive/pull/2483/commits/4c4ced41c8d9b95cc894a5ac091342110006b888#diff-dbfb1120b75fb3b793465548a91c0c40R80
[2]: https://github.com/azurestandard/beehive/pull/2391/files#diff-dbfb1120b75fb3b793465548a91c0c40R54

shofetim:
@wking looks good to me.

PR #190, opened 2017-01-25
Allow specifying a stock when sending test print requests.

caseybessette:
For https://github.com/azurestandard/beehive/issues/2509.

### DIFF
--- a/public.json
+++ b/public.json
@@ -599,7 +599,7 @@
     },
     "/bartender-print-configuration/{id}/actions/test-print": {
       "post": {
-        "summary": "Create a test BarTender print request using this print configuration.  The print request uses hard-coded dummy data.",
+        "summary": "Create a test BarTender print request using this print configuration.  The print request uses hard-coded dummy data if a stock ID is not provided.",
         "operationId": "actionTestBartenderPrintConfiguration",
         "tags": [
           "bartender"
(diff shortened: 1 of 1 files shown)


public.json
```diff
@@ -612,6 +612,20 @@
             "required": true,
             "type": "integer",
             "format": "int64"
+          },
+          {
+            "name": "stock",
+            "in": "body",
+            "description": "Stock data to use in test print request (a payload is required, but the stock is optional)",
```
wking: Maybe:

> Stock data to use in test print request.  The payload must be a JSON object, but the `stock` property is optional.

public.json
```diff
@@ -612,6 +612,20 @@
             "required": true,
             "type": "integer",
             "format": "int64"
+          },
+          {
+            "name": "stock",
```
wking: This is about passing print-request data in, and a stock ID is one way to do that.  To avoid potential confusion with `stock` for the parameter and property names, maybe call this `testData`?

public.json
wking: With a [more generic name on the POSTed object][1], the “payload must be a JSON object…” portion of this description becomes more obviously covered by the JSON Schema itself (declaring `testData` required and not declaring `stock` required).

[1]: https://github.com/azurestandard/api-spec/pull/190/files#r97856389

wking:
I merged this, but fumbled the branch-update order so GitHub thinks it was closed (and doesn't realize it was merged).

PR #192, opened 2017-03-02
API Spec for the Company Directory

tompetersjr:
Backend work in Beehive [#2593](https://github.com/azurestandard/beehive/pull/2593)

### DIFF
--- a/public.json
+++ b/public.json
@@ -745,6 +745,146 @@
         }
       }
     },
+    "/company-directory": {
+      "get": {
+        "summary": "Search company directory",
+        "operationId": "findCompanyDirectoryEntries",
+        "tags": [
+          "company-directory"
+        ],
+        "parameters": [
+          {
+            "name": "query",
+            "in": "query",
+            "description": "Query to perform within the directory.  The use of a `*` at the beginning or end of the string is treated as a wildcard.  This parameter is case insensitive.",
+            "required": false,
+            "type": "string"
+          },
+          {
+            "name": "query-field",
+            "in": "query",
+            "description": "Fields to search.  Any property included in a companyDirectoryEntry response may be queried.  Defaults to `full-name`, `first-name`, `last-name`, `title`, `department.name`, `phone-number`, `phone-ext`, `mobile`, `email`, and `office`.",
+            "required": false,
+            "type": "array",
+            "items": {
+              "type": "string"
+            },
+            "collectionFormat": "csv"
+          },
+          {
+            "name": "department",
+            "in": "query",
+            "description": "Filter by department.  If this parameter is used then department will be ignored in the query-field parameter.  This parameter is case insensitive.",
+            "required": false,
+            "type": "string"
+          },
+          {
+            "name": "sort",
+            "in": "query",
+            "description": "Order results by this field.  Any property included in a companyDirectoryEntry response may be used here.  Defaults to `department.name`, `first-name` and `last-name`.",
+            "required": false,
+            "type": "array",
+            "items": {
+              "type": "string"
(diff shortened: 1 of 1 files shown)


caseybessette:
@tompetersjr I pushed a fixup commit for you to review.  I dropped unnecessary descriptions, requirements (those are needed when POSTing), and listing of fields (just have the user refer to the model definition instead).  Let me know what you think.

tompetersjr:
I'm good with your fixup here.

wking:
On Fri, Mar 03, 2017 at 09:30:45AM -0800, Casey M. Bessette wrote:
> … requirements (those are needed when POSTing)…

‘required’ is not inherently directional.  On a model that is returned
by the API, required tells users what properties they can rely on
Beehive setting.  On a model that is pushed to the API, required tells
Beehive what properties they can require users to set.  So if we're
always going to be setting all of those companyDirectoryEntry
properties, I'm in favor of restoring ‘required’.

public.json
```diff
@@ -6104,6 +6244,70 @@
         "template"
       ]
     },
+    "companyDirectoryEntry": {
+      "description": "a company directory entry",
+      "type": "object",
+      "properties": {
+        "full-name": {
```
wking: There's a lot of overlap between this model and our existing [`person`][1], [`address`][2], and [`telephone`][3] models.  The backend will have to do a bit of shuffling around to make this happen, but I'd prefer a response model like:

```json
{
  "person": {
    "$ref": "#/definitions/person"
  },
  "address": {
    "$ref": "#/definitions/address"
  },
  "telephone": {
    "description": "The employee's telephone numbers",
    "type": "array",
    "items": {
      "$ref": "#/definitions/telephone"
    }
  },
  "title": {
    "type": "string"
  },
  "department": {
    "$ref": "#/definitions/department"
  },
  "office": {
    "type": "string"
  } 
}
```

That's not quite right, because we wouldn't set `person.id` (we could maybe use `updatePerson`, or create a new `person` subset to just drop `id`).

Also, is `office` a room number?  If so, it should probably be part of the `address`.  Maybe just put it in `extended-address`?

[1]: https://github.com/azurestandard/api-spec/blob/fd645a33d8e3501225c73d2c75a3a994b0ed9417/public.json#L8784-L8838
[2]: https://github.com/azurestandard/api-spec/blob/fd645a33d8e3501225c73d2c75a3a994b0ed9417/public.json#L8946-L8987
[3]: https://github.com/azurestandard/api-spec/blob/fd645a33d8e3501225c73d2c75a3a994b0ed9417/public.json#L8750-L8774

tompetersjr: This is all data directly from Microsoft Active Directory (MS AD).  It has no correlation with the Beehive models.  At this time we are not updating MS AD through the API.

wking: > It has no correlation with the Beehive models.

Right, but consuming the API is more straightforward if you don't have to think “is that the address that uses `zip` or the one that uses `postal-code`?”.  It's also easier to convert to formats like [vCard][1] if we return results that are closer to their modeling, and our `telephone` is very close to their [`TEL`][2].  These aren't major issues, and we can always adjust the API later if we want to reduce the overlap, but reducing overlap out of the gate, in situations where it's easy to do, avoids the cost of updating clients if we decide to reduce overlap later on.

[1]: https://tools.ietf.org/html/rfc6350
[2]: https://tools.ietf.org/html/rfc6350#section-6.4.1

caseybessette: We are going to merge this and make the API responses more consistent with the rest of the API when we develop consumers for this (there aren't any planned at this point).

PR #194, opened 2017-04-06
Add product.short-description to definition.

caseybessette:
For https://github.com/azurestandard/beehive/pull/2691.

### DIFF
--- a/public.json
+++ b/public.json
@@ -7033,8 +7033,12 @@
           "description": "sentence-length product name",
           "type": "string"
         },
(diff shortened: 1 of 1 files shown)


wking:
I've pushed a fixup commit that tries to clarify the relationship
between description and short-description as well as describing their
markup language (or lack thereof).  Let me know if it looks good, or
if it needs further tweaks before squashing and merging.

caseybessette:
Your fixup commit looks good to me.

PR #195, opened 2017-05-08
public.json: Remove email confirmation

wking:
In today's meeting, David Stelzer said that prompting users to confirm their email with warnings like “we won't be able to send cutoff reminders until you confirm your email address” will be insufficient motivation.  Instead, we should assume addresses work as entered, and only record errors for bounces and such.  That means we don't need a
confirmation URL in the registration email, since there hasn't been time for anything to go wrong yet.

This is somewhat unwinding #183, although I've left the more-specific `authentication-base-url` (instead of changing it back to `base-url`) to protect against other sorts of URLs we might want to include later.

### DIFF
--- a/public.json
+++ b/public.json
@@ -5137,39 +5137,6 @@
         ],
         "responses": {
           "200": {
-            "description": "Email response.  An email with a confirmation URL will be mailed to the address, and new emails will be unconfirmed until the token is submitted via emailConfirmation.",
-            "schema": {
-              "$ref": "#/definitions/updatedEmail"
-            }
-          },
-          "default": {
-            "description": "Unexpected error.",
-            "schema": {
-              "$ref": "#/definitions/errorModel"
-            }
-          }
-        }
-      }
-    },
-    "/emails/confirm": {
-      "post": {
-        "summary": "Complete an in-progress email confirmation.",
-        "operationId": "emailConfirmation",
-        "tags": [
-          "email"
-        ],
-        "security": [],
-        "parameters": [
-          {
-            "name": "token",
-            "in": "body",
-            "description": "The confirmation token from addEmail or updateEmail.",
-            "required": true,
-            "type": "string"
-          }
-        ],
-        "responses": {
-          "200": {
             "description": "Email response.",
             "schema": {
               "$ref": "#/definitions/email"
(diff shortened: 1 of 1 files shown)


wking:
While working up test notes for azurestandard/beehive#2769, I noticed that once we assume working emails on registration, there's not much point to email confirmation at all.  I've pushed 2e8089b → 870a28a to drop the remaining email-confirmation bits, although this means we'll need to synchronize some of the deploy with frontend changes.

davidmcatee-azure:
Does this mean we are dropping the ability for users to clear any email error?

wking:
On Thu, Aug 24, 2017 at 07:56:50PM +0000, David McAtee wrote:
> Does this mean we are dropping the ability for users to clear any
> email error?

No, they will be able to use PUT to place whatever they want in
email.error or remove email.error entirely without needing
confirmation.  I've pushed 870a28a → be9e784 updating
updateEmailUserPassword to use the email model directly as part of
that which I'd missed before.

davidmcatee-azure:
So would the button on the UI be changed from "Confirm Email" to "Clear This Error"?

wking:
On Thu, Aug 24, 2017 at 08:20:30PM +0000, David McAtee wrote:
> So would the button on the UI be changed from "Confirm Email" to
> "Clear This Error"?

Yeah.  Or drop the button entirely and add an edit modal where the
address and error are editable (or some edit UI that gives you space
to capture the required updateEmailUserPassword.user-password).

caseybessette:
Email addresses in the database are unique per person.  So two people can have the same address.  Maybe I'm missing something.  Won't a malicious user be able to add a secondary email address to his account (without confirming it), make it the primary email, log out, and then request a password reset?  What happens in that case?

wking:
On Fri, Aug 25, 2017 at 08:12:42AM -0700, Casey M. Bessette wrote:
> Won't a malicious user be able to add a secondary email address to
> his account (without confirming it), make it the primary email, log
> out, and then request a password reset?

Hmm, good point.  And User.email is not unique either.  So that
workflow will certainly be possible.  However, I'm not sure it's much
different from what we have now, where you can pass another person's
address to the /password/reset endpoint.  We'll send reset messages
for all users who share that email/username/customer-number [1] and
then on confirmation get the user ID from the token [2] and validate
that user by comparing passwords [3].

The main difference I see with allowing unconfirmed addresses is that
it's easier to spam other users with more than account-creation and
password-reset email.  If we get complaints about that, we'll have to
figure out a new way to handle it.  But I don't see any
account-takeover workflows yet.

[1]: https://github.com/azurestandard/beehive/blob/47b1a54fd5676ef146a4d4bde3e3928eee020022/apps/api/views/password.py#L27-L46
[2]: https://github.com/azurestandard/beehive/blob/47b1a54fd5676ef146a4d4bde3e3928eee020022/apps/api/views/password.py#L70
[3]: https://github.com/azurestandard/beehive/blob/47b1a54fd5676ef146a4d4bde3e3928eee020022/apps/api/views/password.py#L79

PR #196, opened 2017-05-23
public.json: Add stop.shipping-fee

wking:
So users can calculate shipping fees will be without changing the order and hitting `/order-fees`.  The backend [currently hard-codes this at 8.5%][1] and allows the setting to be toggled at the [route-stop][2] level [with a route-level mask][3], so I have not added the new property to `updateStop`.

[1]: https://github.com/azurestandard/beehive/blob/3f30eac07f26dfa62785f4e6999d5562adafbba0/apps/drops/models.py#L983-L984
[2]: https://github.com/azurestandard/beehive/blob/3f30eac07f26dfa62785f4e6999d5562adafbba0/apps/drops/models.py#L839
[3]: https://github.com/azurestandard/beehive/blob/3f30eac07f26dfa62785f4e6999d5562adafbba0/apps/routes/models.py#L206-L210

### DIFF
--- a/public.json
+++ b/public.json
@@ -6770,6 +6770,9 @@
           "description": "offset from midnight on the morning of the trips's delivery start until the usual delivery time for this drop",
           "type": "string",
           "format": "duration"
+        },
+        "drop-shipping-fee": {
+          "$ref": "#/definitions/dropShippingFee"
         }
       },
       "required": [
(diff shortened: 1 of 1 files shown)


wking:
After discussion today, we're going to hang this on the route-(drop|stop) association (#197) and eventually freeze it on the order(fee) to handle changes (azurestandard/beehive#2812).

wking:
Rebased to use `drop-shipping-fee` in c68c8b1 → eaa27c0.

PR #197, opened 2017-05-23
Add routeDropAssociation definition and drop inline option.

caseybessette:
And the dropShippingFee definition.  This is used to expose data for
calculating the shipping fee for drop shipments.

### DIFF
--- a/public.json
+++ b/public.json
@@ -692,6 +692,17 @@
             "collectionFormat": "csv"
           },
           {
+            "name": "inline",
+            "in": "query",
+            "description": "for convenience, include a full representation of the inlined property. Currently supports \"route-drop-associations\".",
+            "required": false,
+            "type": "array",
+            "items": {
+              "type": "string"
+            },
+            "collectionFormat": "csv"
+          },
+          {
             "name": "sort",
             "in": "query",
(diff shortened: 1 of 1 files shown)


public.json
```diff
@@ -6630,6 +6663,28 @@
     "newRoute": {
       "$ref": "route"
     },
+    "routeDropAssociation": {
```
wking: We should use the existing [`routeStop`][1] for this.

[1]: https://github.com/azurestandard/api-spec/blob/6d97a87eeb6b5445924f705e734aba1b12220710/public.json#L6752-L6782

public.json
```diff
@@ -692,6 +692,17 @@
             "collectionFormat": "csv"
           },
           {
+            "name": "inline",
+            "in": "query",
+            "description": "for convenience, include a full representation of the inlined property. Currently supports \"route-drop-associations\".",
```
wking: I think the inline slug should be `route-stops` to match [our currently specified but not implemented endpoint][1].

[1]: https://github.com/azurestandard/api-spec/blob/6d97a87eeb6b5445924f705e734aba1b12220710/public.json#L1718

caseybessette:
We decided to go with https://github.com/azurestandard/api-spec/pull/196 instead.

wking:
Closed in favor of #196, because that API will work even after a drop is removed from a route.

PR #201, opened 2017-06-15
WIP: public.json: Add /person/{id}/prices for requesting per-customer prices

wking:
The current retail/wholesale/member categorization is too coarse, and the business folks want a way to have per-customer deals for customers (or groups of customers) on packaged products (or groups of packaged products).  With this commit, we remove the old price-levels and replace them with a single `basePrice` for each packaged-product.  Individual customers (or customer service representatives acting on behalf of customers) can hit the new `/prices` endpoint to find packaged-products where they have a custom price that's not the default.  This commit covers use cases like “all employees get 10% off everything”, “all members of drop 12 get $5 off food→flour packaged-products”, and “customer 34 gets $3 off FL006”.

One future use case might be promotional codes, where users who push a promo code (to an as-yet-unspecified API endpoint) would pick up the reduced price.

Caching this sort of thing is going to be tricky, because the parameters that a custom price depends on are unclear.  Based on a dev meeting with @tompetersjr, @shofetim, @caseybessette, and @davidmcatee-azure today, we're expecting prices returned by this endpoint to generally be valid until midnight, which we can represent with the usual cache-control HTTP headers.  It's not clear to me yet if having a custom price returned to you via the API guarantees you at least that price until the cache expires.  For example,

1. You hit /prices and are quoted $10 for FL006 (because you're a member of drop 12, although the API doesn't tell you this).  The default price for FL006 is $12.
2. You leave drop 12.
3. You add an order-line for FL006 to your placed order scheduled to be delivered to drop 34.

Can the backend give you the $12 per-unit price on your order-line?  Must it remember that you had a valid-until-midnight $10 quote and give you that instead?  Will we never adjust custom-price logic on the backend in ways that might violate a claimed valid-until-midnight cache statement (this might be tricky if we base prices on the current, possibly estimated, landed cost)?

The query filtering is by product instead of packaged-product because users are expected to always want to know the prices (custom or otherwise) for all packaging associated with a given product.  The frontend may display this information as a price range (or a price-per-unit range) even if the customer is not shown a separate price for each packaging.

FIXME: Will we have any products where we are going to price them by weight, but the MSRP is fixed (not by weight) or vice versa?  Or will we always agree with the manufacturer about whether a given packaged-product should be by weight or not?

FIXME: What values do we want per-unit numbers for?  MSRP-per-unit?  Azure-default-retail-per-unit?  Your-customer-specific-price-per-unit?

FIXME: Would we ever want different wholesale/retail MSRPs?

FIXME: Do we need a way to tag sale items for Algolia filtering?  The price-level-based `on-sale-*` tags I'm removing in this commit currently back https://www.azurestandard.com/shop/tag/on-sale.  What are we doing without that page now that “what products are on sale?” becomes a per-customer question?  Do all “deals” count as “sales”, or are “sales” a subset of “deals”, or are “sales” and “deals” completely independent things?

Fixes azurestandard/beehive#2864.

### DIFF
--- a/public.json
+++ b/public.json
@@ -5087,6 +5087,78 @@
         }
       }
     },
+    "/person/{id}/prices": {
+      "get": {
+        "summary": "Returns all prices from the system for a given person",
+        "operationId": "findPersonPricesById",
+        "tags": [
+          "person",
+          "price"
+        ],
+        "parameters": [
+          {
+            "name": "id",
+            "in": "path",
+            "description": "ID of person whose prices are being requested.",
+            "required": true,
+            "type": "integer",
+            "format": "int64"
+          },
+          {
+            "name": "product",
+            "in": "query",
+            "description": "Product IDs to filter by.",
+            "required": false,
+            "type": "array",
+            "items": {
+              "type": "string"
+            },
+            "collectionFormat": "csv"
+          },
+          {
+            "name": "limit",
+            "in": "query",
+            "description": "Maximum number of results to return.",
+            "required": false,
+            "type": "integer",
+            "format": "int32"
+          },
+          {
+            "name": "start",
+            "in": "query",
+            "description": "Offset to the first result to return.  Use negative numbers to offset from the end of the result list.",
+            "required": false,
+            "type": "integer",
+            "format": "int32"
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "Price response.",
+            "schema": {
+              "type": "array",
+              "items": {
+                "$ref": "#/definitions/price"
+              }
+            },
+            "headers": {
+              "Count": {
+                "description": "Total number of matching results (how many you'd get if you could set an infinite `limit`).",
+                "type": "integer",
+                "format": "int32",
+                "minimum": 0
+              }
+            }
+          },
+          "default": {
+            "description": "Unexpected error.",
+            "schema": {
+              "$ref": "#/definitions/errorModel"
+            }
+          }
+        }
+      }
+    },
     "/session": {
       "get": {
         "summary": "Returns information about the current cookie-based session",
@@ -6928,7 +7000,7 @@
           "format": "float"
         },
         "price": {
-          "#ref": "#/definitions/price"
+          "#ref": "#/definitions/priceBase"
         },
         "stock": {
           "description": "amount of stock available for purchase.  This includes breakdown ancestors.  For example, if we have a single 4 x 6 x 14 oz case in the warehouse, the 4 x 6 x 14 oz packaging will have 1 stock, the 6 x 14 oz packagaging will have 4 stock, and the 14 oz packaging will have 24 stock.",
@@ -7002,7 +7074,6 @@
           "type": "string",
           "enum": [
             "status",
-            "price-level",
             "characteristic"
           ]
         }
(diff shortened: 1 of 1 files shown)


public.json
```diff
@@ -5812,6 +5812,77 @@
...
+      "get": {
+        "summary": "Returns all prices from the system for a given person",
+        "operationId": "findPrices",
+        "tags": [
+          "price"
+        ],
+        "parameters": [
+          {
+            "name": "id",
+            "in": "path",
```
caseybessette: Nit: I'd prefer the person ID in the path, but this is just a nit so do as you please. :)

public.json
```diff
@@ -7188,77 +7258,75 @@
...
+    },
+    "price": {
+      "description": "Customer-specific packaged-product price information.",
       "type": "object",
       "properties": {
-        "dollars": {
-          "description": "price of the product in dollars",
-          "type": "number",
-          "format": "float"
+        "packaged-product": {
```
caseybessette: Since the product ID is used to get the prices, and multiple IDs can be sent in, it makes sense here to include the product ID in the response (in addition to the packaged-product code).

public.json
```diff
@@ -7188,77 +7258,75 @@
...
-          "description": "the per-unit price",
-          "type": "number",
-          "format": "float"
+        "max": {
+          "description": "The maximum orderLine.quantity-ordered (inclusive) for which this price applies.  If max is unset, there is no upper bound.  For example, if max is 3, the price will apply to an order line if quantity-ordered is 2 or 3 but not if quantity-ordered is 4 or 5.",
+          "type": "string"
         },
-        "unit": {
-          "description": "The unit for the dollars-per-unit",
+        "description": {
```
caseybessette: How is `description` going to work if there is more than one stackable deal combined to create the lowest price?  Perhaps this should be an array.

public.json
```diff
@@ -7188,77 +7258,75 @@
...
-          "type": "number",
-          "format": "float"
+        "packaged-product": {
+          "description": "Packaged product code that this price applies to.",
+          "type": "string"
         },
-        "per-pound": {
-          "description": "whether the 'dollars' price is per-pound or per-product",
-          "type": "boolean"
+        "dollars": {
```
caseybessette: Perhaps `dollars` and `dollars-per-unit` below should be `price` and `price-per-unit` since the other values (`default` and `msrp`) are also given in dollars.

public.json
wking: > Nit: I'd prefer the person ID in the path…

That's what we have here (with `"in": "path"`).  Or are you suggesting `/person/{id}/prices` instead of the current `/prices/{id}`?

caseybessette: Oops, I meant to say for the person ID to be in the query string.  `/person/{id}/prices` would be another alternative to accomplish the same thing.

public.json
wking: > Since the product ID is used to get the prices, and multiple IDs can be sent in, it makes sense here to include the product ID in the response…

I left it out of the initial round because it takes up space (if we're trying to optimize bandwidth) and in (almost?) every place you use this object, you'll already have product ID ↔ packaging code information locally.  For example, if you're showing search results, you'll have the product ID and packaging codes from the Algolia response.

If we decide that we do want to include the product IDs and aren't worried enough about payload sizes, I'd rather use per-product objects:

```
[
  [
    "id": 6835,
    "packaging": [
      {
        "code": "FL006",
        …
      },
      …
    ]
  ],
  …
]
```

Or if we drop the usual array-based plural approach, we could key by the IDs and codes:

```
{
  "6835": {
    "FL006": {
        …
    },
    …
  },
  …
}
```

public.json
wking: > Perhaps `dollars` and `dollars-per-unit` below should be `price` and `price-per-unit`…

Yeah.  I was leaving it alone until we addressed the [“What values do we want per-unit numbers for?” FIXME][1].  We might need multiple `*-per-unit` entries.

[1]: https://github.com/azurestandard/api-spec/pull/201#issue-236326755

public.json
wking: > `/person/{id}/prices` would be another alternative to accomplish the same thing.

The person ID isn't optional, so I'd rather keep it in the path.  I've updated this PR with be68e64 → fe7e1f2 to use `/person/{id}/prices`.

public.json
wking: > Perhaps this should be an array.

Made it an array with fe7e1f2 → 5373cab.

wking:
GitHub is having trouble tracking which sub-threads still apply, but I think we currently have to resolve:

* [Do we want to include product IDs in the response, and if so, how?][1]
* [What prices need `*-per-unit` information?][2]

[1]: https://github.com/azurestandard/api-spec/pull/201#discussion_r122464645
[2]: https://github.com/azurestandard/api-spec/pull/201#discussion_r122466857

wking:
> It's not clear to me yet if having a custom price returned to you via the API guarantees you at least that price until the cache expires.

Based on the current meeting, we are ok not making any guarantees here, and floating all prices until the customer freezes a checkout-time price by creating a line in a placed+ order.

PR #202, opened 2017-06-16
public.json: Add ?inline=ancestors for /categories and /category/{id}

wking:
So that clients can walk back to the root category with a single API request.

Previous discussion in azurestandard/beehive#2785 and azurestandard/beehive#2868.

### DIFF
--- a/public.json
+++ b/public.json
@@ -2933,6 +2933,17 @@
             "collectionFormat": "csv"
           },
           {
+            "name": "inline",
+            "in": "query",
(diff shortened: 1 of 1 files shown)


public.json
```diff
@@ -2933,6 +2933,17 @@
             "collectionFormat": "csv"
           },
           {
+            "name": "inline",
+            "in": "query",
+            "description": "For convenience, include a full representation of the inlined property.  Currently supports \"parent\".",
```
caseybessette: You have `parent` here (and below) as being supported, but I think you mean `ancestors`.  I'll push a fixup commit, squash, and merge.

PR #204, opened 2017-06-20
public.json: Add sale model

wking:
With the coming price restructuring, we're going to be applying sales on the frontend.  This commit lets us do that independent of the other price restructuring.

### DIFF
--- a/public.json
+++ b/public.json
@@ -8199,7 +8199,7 @@
       "type": "object",
       "properties": {
         "dollars": {
-          "description": "price of the product in dollars",
+          "description": "Pre-sale price of the product in dollars.",
           "type": "number",
           "format": "float"
         },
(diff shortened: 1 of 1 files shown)


wking:
I've pushed 5da9edee0 → 907b23b90 going from cent rounding to nickle rounding so we can address the rest of this PR without waiting on azurestandard/beehive#2875.

wking:
I've pushed 907b23b → f208df1 adding sale endpoints (907b23b just had inlined sale models).  This will allow us to use the public sale model for the internal pricing UI as well.

wking:
I've pushed f208df1 → 476ca05 adding some notes about handling multiple active sales (only the largest active sale matters).

wking:
I've pushed 476ca05 → bcb9755, rebasing onto master and restoring the cent-rounding (previously removed [here][1]), now that it sounds like we're going ahead with the nickle-round drop.

Also ping @caseybessette on this PR as a whole, since you've been working on the phase-III pricing that this is a part of.

[1]: https://github.com/azurestandard/api-spec/pull/204#issuecomment-309918507

caseybessette:
We have decided to take a different implementation route.

PR #205, opened 2017-06-21
public.json: Add /files and /file/{id}

wking:
The way we've been managing product images internally has proven
awkward and buggy.  This commit adds a file API so we can handle
images (and other opaque files) more explicitly and benefit from the
same logging that applies to the rest of our API.

There is no GET support for individual files, because the contents are
available via https://media.azurestandard.com/files/{id}.

There is no PUT support for individual files, because files are
immutable to support more efficient caching.  If you want to replace a
product or category image, you should:

1. Upload the new image via POST /files, remembering the resulting
   file ID.
2. PUT to the category or packaged-product with the new file ID.
3. DELETE /file/{old-id}.


wking:
Rerolled with 77689af → 79ad3e8 to add a new `packagedProductImage` model.  This makes editing images much more straightforward than having them embedded withing the complicated `packagedProduct` model, but consumers will have to do a bit more legwork to recover packaged-product image URLs.  For example:

```
$ echo '{"images": [{"id": 12, "packaged-product": "FL006", "image": "aa-bb-cc", "order": 3}]}' | jq '[.images[] | "https://media.azurestandard.com/file/" + .image]'
[
  "https://media.azurestandard.com/files/aa-bb-cc"
]
```

PR #208, opened 2017-07-17
public.json: Add ?name=... to /brands

wking:
The internal pricing UI wants to use this for autocompletion.  We could update the pricing UI to use Algolia, but implementing this filter on our backend is pretty easy, and it will let the pricing UI work even without external connectivity.

### DIFF
--- a/public.json
+++ b/public.json
@@ -650,6 +650,13 @@
         ],
         "parameters": [
           {
+            "name": "name",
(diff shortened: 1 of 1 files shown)


wking:
I've also added a commit removing auth for the brand endpoints.  I'm pretty sure we wanted that in #207 and just overlooked it.

PR #211, opened 2017-08-10
Fields for Post Affiliate Pro (PAP)

tompetersjr:
Adds the following fields to the /person endpoint

- affiliate-code
- is-allowed-to-be-affiliate
- accepted-affiliate-terms

Add the following fields to the /registration endpoint

- pap-visitor
- referring-affiliate
- pap-banner

Supporting [Beehive #2986](https://github.com/azurestandard/beehive/pull/2986)

### DIFF
--- a/public.json
+++ b/public.json
@@ -8614,6 +8614,9 @@
           "description": "Redirect URL for the channel-confirmation UI.  The channel-confirmation notification points customers at {confirmation-base-url}{confirmation-token} to confirm the notification channel.",
           "type": "string",
           "format": "url"
+        },
+        "post-affiliate-pro": {
+          "$ref": "#/definitions/postAffiliatePro"
         }
       },
       "required": [
@@ -8630,6 +8633,28 @@
         "registration",
         "checkout",
         "catalog-request"
+      ]
+    },
+    "postAffiliatePro": {
+      "description": "Business logic around the affiliate program is handled by Beehive.  If Beehive determines the referrer is due credit then this information is passed on to Post Affiliate Pro as a sale.",
+      "type": "object",
+      "properties": {
+        "referrer": {
+          "description": "Post Affiliate Pro referring affiliate code.",
+          "type": "string"
+        },
+        "visitor": {
+          "description": "Post Affiliate Pro visitor id.  This information is opaque to Azure.",
+          "type": "string"
+        },
+        "banner": {
+          "description": "Post Affiliate Pro banner ad id.  This information is opaque to Azure.",
+          "type": "string"
+        }
+      },
+      "required": [
+        "referrer",
+        "visitor"
       ]
     },
     "registrationResponse": {
@@ -8839,6 +8864,18 @@
         "internal-notes": {
           "description": "free-form Markdown notes for any internal information that doesn't fit into an existing field.  Only Azure employees can view or edit this information, although it may be passed into the registerPerson endpoint when registering new people.",
           "type": "string"
+        },
+        "affiliate-code": {
+          "description": "Affiliate code to be used with URL shares.",
+          "type": "string"
+        },
+        "is-allowed-to-be-affiliate": {
+          "description": "Is allowed to be an affiliate.",
+          "type": "boolean"
+        },
+        "accepted-affiliate-terms": {
+          "description": "Has accepted the affiliate terms of service.  This can only be set to true if is-allowed-to-be-affiliate value is true.",
+          "type": "boolean"
         }
       },
       "required": [
@@ -8876,6 +8913,18 @@
         },
         "notifications": {
           "$ref": "#/definitions/personNotifications"
+        },
+        "affiliate-code": {
+          "description": "Affiliate code to be used with URL shares.  Only users with the update-person permission can edit this information.",
+          "type": "string"
+        },
+        "is-allowed-to-be-affiliate": {
+          "description": "Is allowed to be an affiliate.  Only users with the update-person permission can edit this information.",
+          "type": "boolean"
+        },
+        "accepted-affiliate-terms": {
+          "description": "Has accepted the affiliate terms of service.  This can only be set to true if the is-allowed-to-be-affiliate value is true.",
+          "type": "boolean"
         }
       }
     },


public.json
```diff
@@ -8467,6 +8467,14 @@
           "description": "Redirect URL for the channel-confirmation UI.  The channel-confirmation notification points customers at {confirmation-base-url}{confirmation-token} to confirm the notification channel.",
           "type": "string",
           "format": "url"
+        },
+        "pap-visitor-id": {
```
wking: We generally don't use `-id` suffixes, even where the property contains an ID.  For example, see `registrationRequest`'s [existing `drop` property][1].  I'd rather use `pap-visitor` here and `referring-affiliate` below.

[1]: https://github.com/azurestandard/api-spec/blob/d11bdbae5f328f4698a2c070081c0b5d025f3f40/public.json#L8600-L8604

public.json
```diff
@@ -8692,6 +8700,18 @@
         "internal-notes": {
           "description": "free-form Markdown notes for any internal information that doesn't fit into an existing field.  Only Azure employees can view or edit this information, although it may be passed into the registerPerson endpoint when registering new people.",
           "type": "string"
+        },
+        "affiliate-code": {
+          "description": "Affiliate code to be used with URL shares",
+          "type": "string"
+        },
+        "is-affiliate": {
+          "description": "Is allowed to be an affiliate",
```
wking: I'm not clear on the distinction between “is” and “is allowed”.  The Beehive PR [suggests][1] that this is about gating a UI and will be set true before the user can view/accept the terms and receive a code.  In that case, maybe `potential-affiliate` would be a more appropriate name.

Also, some words about how this gets set would be helpful.  We don't want to paint ourselves into a corner by being too specific, but there might be something general we an say about when this gets set.

[1]: https://github.com/azurestandard/beehive/pull/2986/files#diff-b7a7d5cb32e1586c14b029ef7b929c92L215

public.json
```diff
@@ -8692,6 +8700,18 @@
...
+          "type": "boolean"
+        },
+        "accepted-affiliate-terms": {
```
wking: This probably needs to go in `updatePerson` as well.

public.json
davidmcatee-azure: @wking, would `is-allowed-to-be-affiliate` be too verbose?  I think it captures the fact that it is something that needs to be granted from above, so to speak. The description could also indicate as such.

wking: > … would is-allowed-to-be-affiliate be too verbose?

I'm fine with verbose, and +1 to documenting the intention in the description.

public.json
```diff
@@ -8839,6 +8851,18 @@
...
+        "affiliate-code": {
+          "description": "Affiliate code to be used with URL shares.  Only Azure employees can edit this information.",
+          "type": "string"
+        },
+        "is-allowed-to-be-affiliate": {
+          "description": "Is allowed to be an affiliate.  Only Azure employees can edit this information.",
+          "type": "boolean"
+        },
+        "accepted-affiliate-terms": {
+          "description": "Has accepted the affiliate terms of service.  This can only be set to true if the is-allowed-to-be-affiliate value is true.",
```
wking: Minor copy-edit: “the is-allowed-to-be-affiliate value” → “is-allowed-to-be-affiliate”.

public.json
```diff
@@ -8839,6 +8851,18 @@
         "internal-notes": {
           "description": "free-form Markdown notes for any internal information that doesn't fit into an existing field.  Only Azure employees can view or edit this information, although it may be passed into the registerPerson endpoint when registering new people.",
           "type": "string"
+        },
+        "affiliate-code": {
+          "description": "Affiliate code to be used with URL shares.  Only Azure employees can edit this information.",
+          "type": "string"
+        },
+        "is-allowed-to-be-affiliate": {
+          "description": "Is allowed to be an affiliate.  Only Azure employees can edit this information.",
```
wking: Editing is only relevant to `updatePerson`, so I think we should drop the “Only Azure…” sentence here  (and from `affiliate-code` above).  We should keep those lines as they stand in the `updatePerson` model below.

public.json
```diff
@@ -8876,6 +8900,18 @@
         },
         "notifications": {
           "$ref": "#/definitions/personNotifications"
+        },
+        "affiliate-code": {
+          "description": "Affiliate code to be used with URL shares.  Only Azure employees can edit this information.",
```
wking: Potentially replace “Azure employees” with “users with the `update-person` permission” (which we already expose in `Person.permissions`) here and below to be more specific.

public.json
```diff
@@ -8614,6 +8614,18 @@
...
+        "pap-visitor": {
+          "description": "Post Affiliate Pro visitor id.  If this is set then referring-affiliate is also required.",
+          "type": "string"
+        },
+        "referring-affiliate": {
+          "description": "Post Affiliate Pro referring affiliate code.  If this is set then pap-visitor is also required.",
+          "type": "string"
+        },
+        "pap-banner": {
+          "description": "Post Affiliate Pro banner id that referred the customer.",
```
wking: It's not clear to me where `pap-banner` comes from.  Is this for [channel tracking][1]?  That seems like an opaque identifier that we pass on to Post Affiliate Pro and don't store locally.  Looking at your current Beehive implementation, I see fields for `campaign_id`, `banner_id`, `channel_id`, `custom_commission`, and `custom_commission_next_tiers_from_campaign` all of which feel somewhat similar.  Can we add more guidance here about what the value should be?  Or can we make our API more generic and accept an opaque `post-affiliate-pro-data` object (or some such) that we turn around and pass up to a particular PAP API?

[1]: https://support.qualityunit.com/438945-SubID--Channels-tracking

public.json
```diff
@@ -8614,6 +8614,18 @@
           "description": "Redirect URL for the channel-confirmation UI.  The channel-confirmation notification points customers at {confirmation-base-url}{confirmation-token} to confirm the notification channel.",
           "type": "string",
           "format": "url"
+        },
+        "pap-visitor": {
+          "description": "Post Affiliate Pro visitor id.  If this is set then referring-affiliate is also required.",
```
wking: It's not clear to me where `pap-visitor` comes from.  Is this some sort of unique token to mark a particular referral (vs. `pap-banner`, because you could have multiple people follow the same banner)?  If so, how is that unique identifier calculated?  What if you are linked to Azure by one affiliate for creating an account, but by another affiliate for ordering a particular line?  Are there more places that we want to collect this sort of thing?

tompetersjr: At this time the affiliate program is being setup such that it works for referrals.  The first affiliate link or banner you click will be the one linked if you register for an account.  pap-banner is optional and only necessary if you followed a banner link.  The pap-visitor is set by a javascript script we include on the page.  We may want to do a hangout if you want to go over the process more.

public.json
tompetersjr: The pap-banner comes from the query string and is the banner ad id clicked on to bring the customer to the site.  I'll update the description to be a bit more informative.

public.json
tompetersjr: I'll post a fixup with this.

PR #217, opened 2017-09-20
Add report endpoints and definitions.

caseybessette:
For https://github.com/azurestandard/beehive/issues/2948 and https://github.com/azurestandard/beehive/issues/3061.

### DIFF
--- a/public.json
+++ b/public.json
@@ -885,6 +885,298 @@
         }
       }
     },
+    "/reports": {
+      "get": {
+        "summary": "Returns all reports from the system that the user has access to",
+        "operationId": "findReports",
+        "tags": [
+          "report"
+        ],
+        "parameters": [
+          {
+            "name": "query",
+            "in": "query",
+            "description": "Query terms to filter by (currently searches `name` only).",
+            "required": false,
+            "type": "string"
+          },
+          {
+            "name": "category",
+            "in": "query",
+            "description": "Report category IDs to filter by.  `null` may be used to return reports that are uncategorized.",
+            "required": false,
+            "type": "array",
+            "items": {
+              "type": "integer",
+              "format": "int64"
+            },
+            "collectionFormat": "csv"
+          },
+          {
+            "name": "limit",
+            "in": "query",
+            "description": "maximum number of results to return",
+            "required": false,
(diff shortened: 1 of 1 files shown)


public.json
```diff
@@ -885,6 +885,298 @@
...
+    "/custom-reports": {
+      "get": {
+        "summary": "Returns all custom reports from the system that the user has access to",
+        "operationId": "findCustomReports",
+        "tags": [
+          "custom-report"
+        ],
+        "parameters": [
+          {
+            "name": "limit",
```
wking: Minor nit, but the existing endpoints usually list `limit` and `start` last.  They're on (almost?) every plural endpoint, so they aren't very interesting.  The parameters which are specific to this endpoint should get the higher-vis positions early in the `parameters` array.

public.json
```diff
@@ -885,6 +885,298 @@
         }
       }
     },
+    "/custom-reports": {
```
wking: Does this endpoint need `custom-`?  I don't expect non-custom reports to ever have an API presence, which means we could use `report(s)` in the API without confusion (or a need to explain that there may also be other kinds of reports).

public.json
```diff
@@ -6709,6 +7001,51 @@
...
+        "id": {
+          "type": "integer",
+          "format": "int64"
+        },
+        "name": {
+          "description": "The category name.",
+          "type": "string"
+        },
+        "customer-pricing-group": {
+          "description": "Whether or not reports in this category are customer pricing groups (for use in the pricing module).",
```
wking: nit: this could be more compact with `custom-pricing` or `customer-pricing`.  I don't expect other booleans around this idea (non-group customer pricing?), so I don't think `-group` is needed.

public.json
```diff
@@ -6709,6 +7001,51 @@
...
+        "id": {
+          "type": "integer",
+          "format": "int64"
+        },
+        "name": {
+          "description": "The name of the report.",
+          "type": "string"
+        },
+        "category": {
+          "description": "The ID of the custom report category the report is in (optional).",
```
wking: No need for `(optional)` (that's represented by `required` below).

PR #221, opened 2017-10-06
public.json: Add person.order-place-issue

wking:
If a customer has carried a sufficiently large balance for a sufficiently long time (as determined by the backend), we want to at least warn users instead of letting them silently checkout (and continue to grow their balance).  We have a few classes of customers for this situation:

* Some customers have fairly serious issues, and they should be encouraged to contact customer service to resolve the issue.  Until then, we will block attempts to place orders.

* Some customers have fairly standard issues, and they should be encouraged to settle their balance directly (e.g. via the “Make a Payment” button [here][1]).  They can also call customer service to sort things out that way.  Until their balance is resolved, we will block attempts to place orders.

* Some customers have special arrangements with Azure to enable them to carry balances while continuing to place new orders.  They should be warned about their outstanding balance, but will still be allowed to place their order.

The enum slugs all start with `outstanding-balance-` to future-proof against other cases where a customer should be blocked or warned from placing all orders.  We don't have those cases now, but using specific slugs leaves room for them if we get some in the future.

The expected workflow is that when users visit a view that allows them to place an order, the frontend will pull a fresh person to get the current value of this property.  If the property is set, the frontend will display an appropriate warning/error, disable the order-placing button, etc.  Because customers are likely to resolve these issues out of band (e.g. via a phone call), frontend warnings should include a suggestion to refresh the data if the customer considers it stale.

There are also related checks that are per-order (vs. per-customer).  For example. we allow customers to place orders after cutoff (but before verification) if their drop received short-drop notifications.  Because they are order-specific, they cannot be represented in this per-person property.  We discussed using `PUT /order/{id}?save=false` for this check, returning errors for these outstanding-balance cases.  That would have allowed us to get the currently-implemented per-order checks at that point as well.  The benefit of the per-customer property is that if at some point we decide to warn customers earlier (e.g. at login, or with a site-wide banner), this approach will work while a per-order approach would require at least an active order.

This property is based @caseybessette's suggestion [here][2].

[1]: https://www.azurestandard.com/my-account/account-balance
[2]: https://github.com/azurestandard/beehive/issues/3062#issuecomment-330681856

### DIFF
--- a/public.json
+++ b/public.json
@@ -5729,7 +5729,7 @@
           {
             "name": "inline",
             "in": "query",
-            "description": "for convenience, include additional information.  Currently supports \"email\" as described in the email model.",
+            "description": "for convenience, include additional information.  Currently supports \"email\" as described in the email model and \"order-place-issue\".",
             "required": false,
             "type": "array",
             "items": {
@@ -5800,7 +5800,7 @@
           {
             "name": "inline",
             "in": "query",
-            "description": "for convenience, include additional information.  Currently supports \"email\" as described in the email model.",
+            "description": "for convenience, include additional information.  Currently supports \"email\" as described in the email model and \"order-place-issue\".",
             "required": false,
             "type": "array",
             "items": {
@@ -5871,6 +5871,19 @@
         "operationId": "findPersonByAuth",
         "tags": [
           "person"
+        ],
+        "parameters": [
+          {
+            "name": "inline",
+            "in": "query",
+            "description": "for convenience, include additional information.  Currently supports \"email\" as described in the email model and \"order-place-issue\".",
+            "required": false,
+            "type": "array",
+            "items": {
+              "type": "string"
+            },
+            "collectionFormat": "csv"
+          }
         ],
         "responses": {
           "200": {
@@ -9373,6 +9386,15 @@
           "items": {
             "$ref": "#/definitions/permission"
           }
+        },
+        "order-place-issue": {
+          "description": "If the person should be warned or blocked from placing orders, this property will be set with a slugged reason and recommended action.",
+          "type": "string",
+          "enum": [
+            "outstanding-balance-contact",
+            "outstanding-balance-pay",
+            "outstanding-balance-warn"
+          ]
         },
         "internal-notes": {
           "description": "free-form Markdown notes for any internal information that doesn't fit into an existing field.  Only Azure employees can view or edit this information, although it may be passed into the registerPerson endpoint when registering new people.",


wking:
We may want to wait on this until [CSR-notification][1] is resolved.  I can see triggering tgat via a faile `PUT /order/{id}?save=false`, but triggering via a `GET /person/{id}` would be more difficult.

[1]: https://github.com/azurestandard/website/issues/1520#issuecomment-334625042

tompetersjr:
+1 on the slug choices.

public.json
```diff
@@ -9374,6 +9374,15 @@
             "$ref": "#/definitions/permission"
           }
         },
+        "order-place-issue": {
+          "description": "If the person should be warned or blocked from placing orders, this property will be set with a slugged reason.",
```
igregson: In addition to **the reason** the person should be warned or blocked from placing orders, this property is also serving to convey: 

- whether they should be warned or blocked
- [maybe] a required action
- [maybe] a suggested action

Accordingly, could we edit this description to read:

"If the person should be warned or blocked from placing orders, this property will be set with a slug that conveys the reason, (maybe) a required action, and (maybe) a recommended action."

public.json
```diff
@@ -9374,6 +9374,15 @@
...
+          "type": "string",
+          "enum": [
+            "outstanding-balance-contact",
+            "outstanding-balance-pay",
+            "outstanding-balance-warn"
```
igregson: In light of my above comment, it seems more intuitive/helpful to have a formula that conveys the full meaning here for creating these slugs. Perhaps: (`[reason]--[warn/block]--[required/suggested action]`)

This would make these:

```
outstanding-balance--block--contact
outstanding-balance--block--pay
outstanding-balance--warn--pay
```

public.json
wking: We don't require frontends do anything.  How about:

> ... with a slugged reason and recommended action.

igregson: Ah, right... yes, that works.

public.json
igregson: If we can go with such a formula, or something similar, it seems wise to document it somewhere (perhaps in the description?).

wking: > If we can go with such a formula, or something similar, it seems wise to document it somewhere (perhaps in the description?).

The current pattern is `{reason}-{frontend-action-hint}`.  I can document that in the commit message, but I'm not sure how it will help API consumers.  If the structure *would* help API consumers, we should use a more structured value than string slugs.

I ecxpext the frontend will take sction for any recognized slug, and show a generic warning for any unrecognized slug.  Being able to patrially-unpack unrecognized entries (e.g. if you understood only the reason part) might allow for more-targetted fallbacks, but I don't expect the backend to so outpace the frontend to be worried about a generic warning.

wking: > The current pattern is {reason}-{frontend-action-hint}. I can document that in the commit message…

Done in d20dfd4 → 3894e5e.

igregson: > The current pattern is {reason}-{frontend-action-hint}. I can document that in the commit message, but I'm not sure how it will help API consumers.

Since the current naming convention used for slugs uses single hyphens between the segments _and within_ the segments, I would prefer we document the pattern (I personally find it helpful, not sure if others would or not). Another option would be to use a naming convention that prevents ambiguity (such as `{reason}-{frontendActionHint}` or `{reason}--{frontend-action-hint}`) and may prevent (or lessen) the need for documenting patterns.

Otherwise, I'm fine with leaving out `block || warn` from the pattern if that's not something that the backend should specify. [The messages to be displayed to the user](https://github.com/azurestandard/website/issues/1520#issuecomment-334622487), however, in all three cases suggestion to contact customer service (leaving ambiguity in the pattern, which `block || warn` would remove).

Just my 2 cents! Since the [issue at hand](https://github.com/azurestandard/website/issues/1520) is labeled `Priority: Critical` I'll go ahead and approve this.

wking: > Another option would be to use a naming convention that prevents ambiguity (such as `{reason}-{frontendActionHint}` or `{reason}--{frontend-action-hint}`) and may prevent (or lessen) the need for documenting patterns.

If you want to act on partially-understood slugs, I suggest we use:

```
"order-place-issue": {
  "reason": "outstanding-balance",
  "action": "pay"
}
```

and in that case we'd document the two fields in the defining Swagger.  Do you expect to switch on those sub-parts independently?

wking:
And I [still][1] think we want to wait on merging this until we know if we'll need to [notify CSRs of warn/block activity][2].  I don't think this API (as it stands in 299e004) would cleanly support that workflow.

[1]: https://github.com/azurestandard/api-spec/pull/221#issuecomment-334628495
[2]: https://github.com/azurestandard/website/issues/1520#issuecomment-334625042

caseybessette:
> We may want to wait on this until [CSR-notification][1] is resolved.  I can see triggering that via a failed `PUT /order/{id}?save=false`, but triggering via a `GET /person/{id}` would be more difficult.
> 
> [1]: https://github.com/azurestandard/website/issues/1520#issuecomment-334625042

If this is indeed required, a possible way to handle this with `PUT /person/id`: Change the enums in this PR to objects that include whether or not a CSR notification should be triggered, and under what condition (customer's reception of notification or customer's action), e.g.,

```javascript
"order-place-issue": {
  "reason": "outstanding-balance-pay",
  "notify-customer-service-on-reception": true  // OR: notify-customer-service-on-action
}
```

Two possible triggers:

1. Provide some UI element that requires action, such as a dialog box with an `Ok` button or an `X` to dismiss, that when clicked will `PUT /person/id` with something like `notifiy-customer-service: true`.  This would need to be blocking (can't use the site without dismissal).  And we would only want to trigger the PUT if it is the customer who is logged in, not superusers.

2. Assume the customer has not seen the notification until the next time the customer logs in.  When the customer (again, not superuser CSRs) logs in and is shown a message that we would want a CSR to proactively act on, we immediately `PUT /person/id` with `notifiy-customer-service: true`.

I was half thinking a database column could be added `notify_customer_service DEFAULT FALSE` so a background task could send CSRs the email, but handling it right in the endpoint works, too.

public.json
igregson: My primary thinking behind a non-ambiguous slug naming convention was for code readability/clarity. As this is, the frontend has what it needs to complete the task.

All the while, the object route here is more robust. 

If we're going to go the object route, however, I'm wondering if we should use something more generic that would be an all-encompassing endpoint for customer/account-specific issues. There's an *account* notification system feature (different than the *site* notifications we currently have) that @davidmcatee-azure and I have discussed some in the past (I thought there was an issue for it but I couldn't find it) that this could be used for. Something along these lines (naming/structure not thoroughly thought out, this is to convey the idea):

```
"issues": [{
  "type": "outstanding-balance",
  "blockCheckout": True,
  "suggestedAction": "contact"
}]
```

All the while, this is

igregson: Given the priority level, however, I'm fine going with this PR as it currently stands and punting on anything more robust for now.

tompetersjr:
I think we should hold on the notifications for now.  It is important that this goes out ASAP and the notification part can happen after the fact.

PR #222, opened 2017-10-13
Add orderParcelCarrierEstimate.contact-for-quote.

caseybessette:
And remove orderParcelCarrierEstimate.message which was from a previous
iteration of the feature specification.

### DIFF
--- a/public.json
+++ b/public.json
@@ -8804,10 +8804,29 @@
           "description": "The estimated latest delivery date (inclusive) for the covered order-lines.",
           "type": "string",
           "format": "date"
+        },
+        "contact-for-quote": {
+          "description": "The customer should contact this person or team for a personalized shipping quote for this service.",
+          "type": "object",
+          "properties": {
+            "name": {
+              "description": "Name of the person or team.",
+              "type": "string"
+            },
+            "email": {
+              "description": "Email for the person or team.",
+              "type": "string",
+              "format": "email"
+            },
+            "telephone": {
+              "description": "E.123 telephone number for the person or team.",
+              "type": "string",
+              "format": "telephone"
+            }
+          }
         }
       },
       "required": [
-        "cost",
         "climate",
         "service"
       ]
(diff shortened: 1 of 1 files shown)


public.json
```diff
@@ -8805,9 +8805,23 @@
...
+          "description": "The customer should contact this person or team for a personalized shipping quote for this service.",
+          "type": "object",
+          "properties": {
+            "name": {
+              "description": "Name of the person or team.",
+              "type": "string"
+            },
+            "email": {
+              "description": "Email of the person or team.",
+              "type": "string"
```
wking: This should also get a `"format": "email"`.  For example:

```
$ jq .definitions.email.properties.address public.json 
{
  "type": "string",
  "format": "email"
}
```

public.json
```diff
@@ -8805,9 +8805,23 @@
...
+            },
+            "phone": {
+              "description": "Phone number of the person or team.",
+              "type": "string"
```
wking: I so want `contact-for-quote` to be a `"$ref": "#definitions/person"` :p.  It looks like that's still a bit out of reach, because phone numbers are still address properties, and not person properties (azurestandard/beehive#950).  However, I'd at least like to follow the existing patterns as much as possible, which we could do following `registrationRequest`:

```json
"telephone": {
  "description": "Contact telephone number for the person or team (will include at least one 'voice' type).",
  "type": "array",
  "items": {
    "$ref": "#/definitions/telephone"
  }
}
```

or dropping the array:

```json
"telephone": {
  "description": "Contact telephone number for the person or team.",
  "$ref": "#/definitions/telephone"
}
```

Or if we really want to hard-code “voice”, following `definitions.telephone.properties.number`:

```json
"telephone": {
  "description": "E.123 telephone number for the person or team.",
  "type": "string",
  "format": "telephone"
}
```

public.json
caseybessette: Fixed.

caseybessette:
@wking This PR has been updated with your suggestions.

PR #226, opened 2018-04-02
Add product ID to favorite and orderLine definitions.

caseybessette:


caseybessette:
@wking Looks good to me.

PR #229, opened 2018-05-02
public.json: Add RFM, first-order-placed, etc. to person

wking:
[As][2] [requested][1] by @davidcross so the frontend can slot in tailored home page content.  Personally, I'd be happier [if the place times only included orders which remained in the database][3], but that's not my call.

We also discussed using cookies for this information, but decided on person properties so customer service can see the same home page as a customer when they are acting on the customer's behalf (`?superuser-for=...`).

[1]: https://github.com/azurestandard/website/issues/2161
[2]: https://github.com/azurestandard/beehive/issues/3735
[3]: https://github.com/azurestandard/beehive/issues/3735#issuecomment-385823868

### DIFF
--- a/public.json
+++ b/public.json
@@ -9570,6 +9570,16 @@
           "format": "int32",
           "minimum": 1,
           "maximum": 5
+        },
+        "first-order-placed": {
+          "description": "When this customer's first order was placed.  This timestamp does not change once set, even if the initial order is subsequently deleted.",
+          "type": "string",
+          "format": "date-time"
+        },
+        "last-order-placed": {
+          "description": "When this customer's most recent order was placed.  This timestamp includes all open-to-placed transitions, regardless of whether the related order is subsequently deleted.",
+          "type": "string",
+          "format": "date-time"
         }
       },
       "required": [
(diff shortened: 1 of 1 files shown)


davidcross:
As I mentioned to David McAtee: I will still need these values in cookies
so I can pull them within WordPress and Optin Monster modal windows pulled
from outside of AS.  Are these values already available in cookies?


David

---
David Cross
david@davidcross.com
Tel: +1-503-805-1019

tompetersjr:
@davidcross They are not.  Only the RFM values are on the cookies right now.  This particular pull requests is only to document the API request updates for the person endpoint.  The https://github.com/azurestandard/beehive/issues/3735 issue is to add the information to the endpoint.  Sounds like we need a new that issue so that we also add `first-order-placed` and `last-order-placed` to the session endpoint for the cookies.

tompetersjr:
Cross linking the issue broken off from https://github.com/azurestandard/api-spec/pull/229#issuecomment-386168669

https://github.com/azurestandard/beehive/issues/3738

PR #230, opened 2018-05-24
public.json: API for the rewards program

wking:
The API side of azurestandard/beehive#3774.

### DIFF
--- a/public.json
+++ b/public.json
@@ -8700,6 +8700,11 @@
           "description": "before shipping, the customer can use this to associate a payment-method with the order.  After shipping, it contains information about the charged payment",
           "$ref": "#/definitions/payment"
         },
+        "rewards-allocation": {
+          "description": "Rewards allocated for this order.  For open orders, this value has no side-effects.  For placed orders that are not yet shipped, this value will match the associated pending rewards debit amount.  When the order ships, this value does not change, but the associated rewards debit will have its amount reduced if the allocated rewards exceed the dollar balance.  If, instead of shipping, the placed order returns to open status, the associated pending rewards debit is deleted, releasing those rewards to be allocated towards other orders.",
+          "type": "number",
+          "format": "float"
+        },
         "drop": {
           "description": "drop the order is destined for (unset for parcel-carrier orders)",
           "type": "integer",
--- a/public.json
+++ b/public.json
@@ -8375,34 +8375,62 @@
       ]
     },
     "_price": {
-      "description": "a helper object for 'price'",
+      "description": "A helper object for `price`.",
       "type": "object",
       "properties": {
         "dollars": {
-          "description": "price of the product in dollars",
+          "description": "Price of the packaging in dollars.",
           "type": "number",
-          "format": "float"
+          "format": "float",
+          "minimum": 0
         },
         "per-pound": {
-          "description": "whether the 'dollars' price is per-pound or per-product",
+          "description": "Whether the `dollars`, `rewards`, and `rewards-standard` values are per-pound or per-package.",
           "type": "boolean"
         },
         "discount": {
-          "description": "text for the discount (e.g. \"12%\", or \"$3.50\")",
+          "description": "Text for the discount (e.g. \"12%\", or \"$3.50\").",
           "type": "string"
         },
         "dollars-per-unit": {
-          "description": "the per-unit price",
+          "description": "The per-unit price.",
           "type": "number",
-          "format": "float"
+          "format": "float",
+          "minimum": 0
+        },
+        "rewards": {
+          "description": "Rewards earned for this packaging.",
+          "type": "number",
+          "format": "float",
+          "minimum": 0
+        },
+        "rewards-standard": {
+          "description": "Hypothentical rewards earned for this packaging using the standard formula.  This value may be less than, equal to, or greater than `rewards`.",
+          "type": "number",
+          "format": "float",
+          "minimum": 0
+        },
+        "rewards-per-unit": {
+          "description": "Rewards earned per unit.",
+          "type": "number",
+          "format": "float",
+          "minimum": 0
+        },
+        "rewards-per-unit-standard": {
+          "description": "Hypothetical rewards earned per unit using the standard formula.  This value may be less than, equal to, or greater than `rewards-per-unit`.",
+          "type": "number",
+          "format": "float",
+          "minimum": 0
         },
         "unit": {
-          "description": "The unit for the dollars-per-unit",
+          "description": "The unit for the `dollars-per-unit`, `rewards-per-unit`, and `rewards-per-unit-standard`.",
           "type": "string"
         }
       },
       "required": [
-        "dollars"
+        "dollars",
+        "rewards",
+        "rewards-standard"
       ]
     },
     "priceLevel": {
--- a/public.json
+++ b/public.json
@@ -4869,7 +4869,7 @@
     },
     "/account-entries": {
       "get": {
-        "summary": "Returns all credits and debits from the system that the user has access to.  Results are ordered by increasing settled dates, falling back to IDs when multiple entries settled on the same day.",
+        "summary": "Returns all dollar credits and debits from the system that the user has access to.  Results are ordered by increasing settled dates, falling back to IDs when multiple entries settled on the same day.",
         "operationId": "findAccountEntries",
         "tags": [
           "account"
@@ -4906,7 +4906,7 @@
           {
             "name": "balance",
             "in": "query",
-            "description": "add the balance field to returned entries",
+            "description": "Add the `balance` property to returned entries.",
             "required": false,
             "type": "boolean"
           },
@@ -4956,7 +4956,7 @@
     },
     "/account-entry/{id}": {
       "get": {
-        "summary": "Returns a credit or debit based on a single ID",
+        "summary": "Returns a dollar credit or debit based on a single ID",
         "operationId": "findAccountEntryById",
         "tags": [
           "account"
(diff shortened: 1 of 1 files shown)


public.json
```diff
@@ -9292,6 +9412,57 @@
...
+          "type": "integer",
+          "format": "int64"
+        },
+        "person": {
+          "description": "Person associated with the account.",
+          "type": "integer",
+          "format": "int64"
+        },
+        "amount": {
+          "description": "Reward value for the entry (positive for earning rewards, negative for spending rewards).",
```
wking: "positive for credits, negative for debits"

public.json
```diff
@@ -9292,6 +9412,57 @@
...
+          "type": "string",
+          "format": "date-time"
+        },
+        "order": {
+          "description": "The order which produced (when `amount` > 0) or consumed (when `amount` < 0) this entry.",
+          "type": "integer",
+          "format": "int64"
+        },
+        "credit": {
+          "description": "The account-entry credit created by this entry.  This will be set when `amount` < 0 where the referenced credit shows the dollar increase associated with this rewards decrease.",
```
wking: "when `amount` < 0" -> "for reward debits"

public.json
```diff
@@ -9292,6 +9412,57 @@
...
+          "description": "When the entry was settled.  This will be unset for pending debits, and will be set for all settled debits.",
+          "type": "string",
+          "format": "date-time"
+        },
+        "order": {
+          "description": "The order which produced (when `amount` > 0) or consumed (when `amount` < 0) this entry.",
+          "type": "integer",
+          "format": "int64"
+        },
+        "credit": {
```
wking: "credit" -> "account-entry"

public.json
```diff
@@ -5186,6 +5186,126 @@
         }
       }
     },
+    "/rewards": {
+      "get": {
+        "summary": "Returns all reward credits and debits from the system that the user has access to.",
```
wking: Talk about ordering (and also for `/account-entries`, which is currently ordered by increasing `assessed_on`).  This will be ordered by `settled`.  For entries which haven't settled, we can sort by the associated order's pull times?  And put them all at the end?  For example:

settled 2010
settled 2011
...
settled 2018
pending, order pulls 2019
pending, order pulls 2020

wking: Maybe say the pending ones have undefined ordering, and then we can decide between `id` and order-pull time on the backend.

wking: Or specify that we order by settled time, and then fall back to ID (corresponding to the creation time).

public.json
```diff
@@ -9292,6 +9412,57 @@
...
+          "type": "integer",
+          "format": "int64"
+        },
+        "balance": {
+          "description": "Settled reward balance after this reward entry.",
+          "type": "number",
+          "format": "float"
+        },
+        "pending": {
+          "description": "Total pending debits through this reward entry.",
```
wking: This will always be >= 0.  100 settled balance - 75 pending = 25 available.

public.json
```diff
@@ -8673,6 +8717,16 @@
           "description": "before shipping, the customer can use this to associate a payment-method with the order.  After shipping, it contains information about the charged payment",
           "$ref": "#/definitions/payment"
         },
+        "reward-credit": {
+          "description": "After shipping, this is the reward credit earned due to this order.",
+          "type": "integer",
+          "format": "int64"
+        },
+        "reward-debit": {
```
wking: Need to put this into the update-order model for POST and PUT as well.

wking: Alternative would be `POST /rewards`, but to do that you'd end up with pending debits on open orders, and we aren't excited about locking points with open orders.  But the order endpoint is complicated.  We're settling on not `POST /rewards`, because we need to atomically lock the association when you transition into placed status.

Mention that we're deleting pending rewards when we re-open an order (on the transition from drop to parcel-carrier).

We're also going to add a property to the order to store the requested rewards debit, so we can store open-order plans without locking the associated rewards.

We can drop the inlined data from the responses here; that can come from `GET /rewards?order=12,34`.

public.json
```diff
@@ -8392,29 +8392,49 @@
...
       "properties": {
         "dollars": {
-          "description": "price of the product in dollars",
+          "description": "Price of the packaging in dollars.",
           "type": "number",
           "format": "float"
         },
         "per-pound": {
-          "description": "whether the 'dollars' price is per-pound or per-product",
+          "description": "Whether the `dollars` price is per-pound or per-package.",
```
wking: Also mention rewards.

public.json
```diff
@@ -8392,29 +8392,49 @@
...
+          "description": "The per-unit price.",
+          "type": "number",
+          "format": "float"
+        },
+        "rewards": {
+          "description": "Rewards earned for this packaging.  This will be >= `rewards-base`.",
+          "type": "number",
+          "format": "float"
+        },
+        "rewards-base": {
```
wking: Maybe only set this if it was smaller than `rewards`.  Drop the >= comment from the `rewards` if we do this.  The bonus-rewards condition would be "if they're both set, show the bonus".  Drawback is that the knowledge about the API is less obvious than `if (price.rewards > price["rewards-base"])`.

public.json
```diff
@@ -8392,29 +8392,49 @@
...
+          "description": "Rewards earned for this packaging using the standard formula.",
+          "type": "number",
+          "format": "float"
+        },
+        "rewards-per-unit": {
```
wking: Maybe leave these out of round 1?  Come back to them if the frontend ends up needing them.

wking: We might remove these later if it turns out that we don't need them (e.g. no "as high as" rewards display).  Not sure where else we use per-unit pricing (distinct from per-pound variable-weight pricing).

public.json
```diff
@@ -8993,7 +9176,7 @@
       ]
     },
     "accountEntry": {
-      "description": "an entry in a person's acount",
+      "description": "A dollar entry in a person's acount.",
```
caseybessette: `acount` -> `account`

public.json
```diff
@@ -9292,6 +9475,59 @@
         "catalog-request"
       ]
     },
+    "reward": {
+      "description": "An reward entry in a person's acount.",
```
caseybessette: `acount` -> `account`

caseybessette: It is a little confusing to speak of rewards as part of `a person's account` here, and then later speak of `The account-entry credit created by this entry.`

public.json
```diff
@@ -5186,6 +5186,153 @@
...
+            },
+            "collectionFormat": "csv"
+          },
+          {
+            "name": "date-after",
+            "in": "query",
+            "description": "Only return entries with dates after this date (inclusive).",
+            "required": false,
+            "type": "string",
+            "format": "date"
```
caseybessette: Perhaps this should be a datetime to match the backend data type?

public.json
```diff
@@ -9292,6 +9475,59 @@
...
+    "reward": {
+      "description": "An reward entry in a person's acount.",
+      "type": "object",
+      "properties": {
+        "id": {
+          "type": "integer",
+          "format": "int64"
+        },
+        "person": {
+          "description": "Person associated with the account.",
```
davidmcatee-azure: `"Person associated with the reward"` instead?

public.json
wking: > It is a little confusing to speak of rewards as part of `a person's account` here...

I'll just simplify to `"A reward entry."` ;).

wking:
Rerolled with dd43f63 -> eb2d059 addressing the most recent round of review.

public.json
```diff
@@ -8228,29 +8375,55 @@
...
+          "minimum": 0
+        },
+        "rewards": {
+          "description": "Rewards earned for this packaging.",
+          "type": "number",
+          "format": "float",
+          "minimum": 0
+        },
+        "rewards-base": {
+          "description": "Rewards earned for this packaging using the standard formula.  This property will only be set if the value is <= `rewards`.",
```
caseybessette: I thought we were going to just always push this?  I don't think it matters, though, unless @davidmcatee-azure has a preference.

davidmcatee-azure: My preference would be to always include it.

wking: > I thought we were going to just always push this?

I couldn't remember where we ended up (brief notes [here][1]).  I tried to explain the direction I went in the eb2d0591f6b commit message.

> My preference would be to always include it.

I'm ok with that, but if we are going to expose `*-base` in all responses, I think we need a position on whether `rewards < rewards-base` (and similar for other `*-base`) is possible.

By only setting `*-base` forms that are lower, we collapse the = and < cases together (as far as the API consumer can tell), so we don't have to make those calls in the API docs.

[1]: https://github.com/azurestandard/api-spec/pull/230#discussion_r190707815

davidmcatee-azure: The way I see it, `rewards` can be anything. `rewards-base` is just there for consumers to compare the actual rewards with the hypothetical rewards according to the standard rate.  It may help if the `rewards-base` description began with "Hypothetical rewards earned ..."

wking: > `rewards-standard` is how I have always thought about it

I have no preference on "base" vs. "standard"; should I change my strings to use "standard"?

> I have assumed that rewards may be lower in some cases.

If we expect to allow those, I think we need to say so in the description string here.  And if we're comfortable making that statement, I don't have a problem with always setting `rewards-base` (or `rewards-standard`, or whatever).

davidmcatee-azure: `rewards-standard` may be a better name for it.

davidmcatee-azure: I would be comfortable stating that `rewards` could be greater than, equal to, or less than `rewards-standard`.

wking:
I've pushed eb2d059 -> 6074db6 changing `-base` to `-standard` and always setting the standard forms (based on [this discussion][1]).  @caseybessette, @davidmcatee-azure, besides reviewing the JSON Schema updates, can you look over the 6074db6 commit message to check that I'm appropriately summarizing the consensus?

[1]: https://github.com/azurestandard/api-spec/pull/230#discussion_r191887254

caseybessette:
Looks good to me, Trevor.

PR #231, opened 2018-06-22
public.json: Split reward.balance into total-credit and total-debit

wking:
This allows us to show things like "lifetime rewards", so folks can see the total impact of the rewards program on their account.  You can retrieve the previous balance value with `credit-total - debit-total`.

We don't want to do this for `accountEntry`, because we don't expect to break that down for folks (e.g. no "lifetime dollars sent to Azure" UI).

### DIFF
--- a/public.json
+++ b/public.json
@@ -5250,9 +5250,9 @@
             "collectionFormat": "csv"
           },
           {
-            "name": "balance",
+            "name": "totals",
             "in": "query",
-            "description": "Add the `balance` and `pending` properties to returned entries.",
+            "description": "Add the `total-credit`, `total-debit`, and `total-pending` properties to returned entries.",
             "required": false,
             "type": "boolean"
           },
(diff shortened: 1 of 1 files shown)


wking:
Actually, let me flip these around to `total-debit` so they sort together with `jq -S ...`.

wking:
Ok, switched to `total-*` with 7d107bb -> 249308e.

davidmcatee-azure:
"pending" is also a total, so I wonder if "total-" is not necessary or `pending` should be `total-pending`.

wking:
> "pending" is also a total, so I wonder if "total-" is not necessary or pending should be total-pending.

I think `credit` and `debit` (or `credits`, etc.) could be confused with the per-entry `amount` as "not a running total".  I'm less concerned about that for `pending` (and for the previous `balance`).  Do you think the unprefixed `pending` might be confusing (in which case we'd want `total-pending`)?  Or do you think `credits` or some such would not be confusing (in which case we'd want `credits`, etc.).  Or do you think `credits` might be confusing, and that we should use `total-pending` for consistency regardless of whether `pending` might be confusing?

davidmcatee-azure:
Yes, good point. I think prefixing all three with "total-" makes the most sense.

wking:
> I think prefixing all three with "total-" makes the most sense.

Ok.  Done with 249308e -> d73baa8.

PR #237, opened 2018-08-27
Add affiliate-related properties to person.

caseybessette:


public.json
```diff
@@ -9840,9 +9840,17 @@
...
+        "affiliate-campaign": {
+          "description": "The affiliate campaign the person is a participant of.",
+          "type": "string",
+          "enum": [
+            "lead",
+            "community"
+          ]
+        },
+        "marketing-slug": {
+          "description": "A person's marketing slug used, e.g., for the Azure Share program.",
```
davidmcatee-azure: Should this say "A person's unique marketing slug"?

caseybessette: A person can have more than one, but there is only one primary.  I'll update the PR to reflect that.

caseybessette: But, they are also unique. :+1:

PR #241, opened 2018-10-25
Add inline "is-drop-coordinator" prop to person definition

igregson:
For https://github.com/azurestandard/beehive/issues/4077.

### DIFF
--- a/public.json
+++ b/public.json
@@ -9969,7 +9969,7 @@
           "format": "date-time"
         },
         "is-drop-coordinator": {
-          "description": "Is a drop coordinator (inclusive of both primary contact and managers). Only set if the \"inline\" query parameter contains \"is-drop-coordinator\").",
+          "description": "Is a drop coordinator (inclusive of both primary contact and managers). Only set if the \"inline\" query parameter contains \"is-drop-coordinator\".",
           "type": "boolean"
         }
       },
--- a/public.json
+++ b/public.json
@@ -5983,7 +5983,7 @@
           {
             "name": "inline",
             "in": "query",
-            "description": "for convenience, include additional information.  Currently supports \"email\" as described in the email model and \"order-place-issue\".",
+            "description": "for convenience, include additional information.  Currently supports \"email\" as described in the email model, \"order-place-issue\", and \"is-drop-coordinator\".",
             "required": false,
             "type": "array",
             "items": {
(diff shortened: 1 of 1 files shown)


igregson:
@davidmcatee-azure Will request a review from @caseybessette once you've taken a look.

public.json
```diff
@@ -9967,6 +9967,10 @@
           "description": "When this customer's most recent order was placed.  This timestamp includes all open-to-placed transitions, regardless of whether the related order is subsequently deleted.",
           "type": "string",
           "format": "date-time"
+        },
+        "is-drop-coordinator": {
+          "description": "Is a drop coordinator (inclusive of both primary contact and managers). Only set if the \"inline\" query parameter contains \"is-drop-coordinator\".",
```
davidmcatee-azure: I noticed that this property's description and `email` both say `set if the \"inline\" query parameter contains ...` but the `order-place-issue` property's description doesn't.  Maybe we should add that to the `order-place-issue` description.

igregson: Good thinking.

caseybessette:
To future-proof this, perhaps there should be two properties (though I'm not sure the correct names to call them since "manager" is a word we want to avoid):

1. `is-drop-coordinator`
2. `is-drop-manager`

I suppose if "coordinator" is a broader term that incorporates both concepts, then what you have here would be appropriate.

igregson:
> I suppose if "coordinator" is a broader term that incorporates both concepts

That was the thinking that @davidmcatee-azure and I came to in discussion. The [current handling](https://github.com/azurestandard/beehive/blob/master/apps/api/views/drops.py#L187-L202) for the `coordinators` property (an array of drop "managers" and the single "coordinator") on the `orders` object uses the term `coordinators` in the broader sense (i.e. the primary contact and managers) and that's the approach we've been considering for UI text on the frontend (all who have access to "manage" the drop are coordinators, one is the primary contact). Even if we settled on a program name change (such as "Drop Host") the thinking is that this would only be a user facing change and that we'd maintain the most logical approach in the code.

It's worth mentioning that after seeing the `is_drop_manager` column in the `drops_droppointmembership` table we discussed simply having this exposed via the API. We could then simply query the `drop-membership` endpoint to achieve the goal of https://github.com/azurestandard/beehive/issues/4077. This assumes that the primary contact is always a manager (of the drop they're the primary contact). Unfortunately, after testing in Beehive we learned that it is possible to be the primary contact and _not_ be a manager. Given this, we concluded that this route wouldn't work (from the sense of it being a more performant approach for the backend). If we're mistaken for some reason, however, this route is preferable.

PR #243, opened 2018-10-25
Add "last-human-update" prop to order definition

igregson:
For https://github.com/azurestandard/beehive/issues/4062.

### DIFF
--- a/public.json
+++ b/public.json
@@ -8891,6 +8891,10 @@
         "notes": {
           "description": "Warehouse packing instructions",
           "type": "string"
+        },
+        "last-human-update": {
+          "description": "a datetime denoting when the order was last updated by a human (rather than programmatically)",
+          "type": "date-time"
         }
       },
       "required": [


igregson:
@davidmcatee-azure Will request a review from @caseybessette once you've taken a look.

public.json
```diff
@@ -8891,6 +8891,10 @@
         "notes": {
           "description": "Warehouse packing instructions",
           "type": "string"
+        },
+        "last-human-update": {
+          "description": "a datetime denoting when the order was last updated by a human (rather than programmatically)",
```
davidmcatee-azure: We might want to say `the order or its order-lines`.

caseybessette:
Perhaps this should be named `last-api-update` instead.  Humans can touch the order in Beehive but we aren't going to track that.  And the API may programmatically update the order and such changes would be included here.

igregson:
> And the API may programmatically update the order [...]

Hmm... I can't think of any case where this currently happens, any particular example of this in mind?

As much as possible the name should reflect that the update has not happened programmatically, since the catalyst and first use case is DCs filtering out stale orders.

Perhaps one of these:

`last-user-update`?
`last-frontend-update`?

igregson:
> I'm going to program it to mean "the last time any aspect of the order was changed via the API." I won't distinguish between if that was a human, if it came from the front-end, or a superusering CSR.

Alright, then I agree that `last-api-update` makes the most sense. Will update the PR.

PR #244, opened 2018-11-14
Add computed total properties to order endpoint

igregson:
Note that this is the first of our new "computed" concept (distinct from "inline"). To recap the difference (from notes taken via discussion that resulting in the concept): 

Use `inline` for nested properties.
Use `computed` for computed properties (returned JSON nested inside a `computed` property)

Also note that I've used camel case here, as I understood we wanted to do all future API-related development using it (and will at some point convert existing hyphenation to it).

### DIFF
--- a/public.json
+++ b/public.json
@@ -8906,23 +8906,38 @@
           "description": "Optional computed properties included via the \"computed\" parameter.",
           "type": "object",
           "properties": {
-            "totalLinePrice": {
-              "description": "Dollar value for the sum of all orderlines.",
+            "totalItemPriceOrdered": {
+              "description": "Dollar value for the sum of all ordered items (does not include potential fees).",
               "type": "number",
               "format": "float"
             },
-            "totalFees": {
-              "description": "Dollar value for the sum of all order fees.",
+            "totalItemPriceShipped": {
+              "description": "Dollar value for the sum of all shipped items (does not include potential fees).",
               "type": "number",
               "format": "float"
             },
-            "totalWeight": {
+            "totalWeightOrdered": {
               "description": "Total weight (in pounds) of all ordered items.",
               "type": "number",
               "format": "float"
             },
-            "totalVolume": {
+            "totalWeightShipped": {
+              "description": "Total weight (in pounds) of all shipped items.",
+              "type": "number",
+              "format": "float"
+            },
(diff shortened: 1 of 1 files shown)


davidmcatee-azure:
Looks good to me.

public.json
```diff
@@ -8901,6 +8901,32 @@
         "last-api-update": {
           "description": "a datetime denoting when the order or its order-lines were last updated via the API",
           "type": "date-time"
+        },
+        "computed": {
+          "description": "Optional computed properties included via the \"computed\" parameter.",
+          "type": "object",
+          "properties": {
+            "totalLinePrice": {
+              "description": "Dollar value for the sum of all orderlines.",
```
caseybessette: Presumably you want the sum of all ordered lines prior to shipping and sum of all shipped lines after shipping?  That might be worth noting in this description.  Alternatively, I could give you `totalLineOrderedPrice` and `totalLineShippedPrice`.  The same distinction is available for weight and volume, but not for fees.

igregson: Just pushed relevant changes. @davidmcatee-azure do you mind taking another look at this?

davidmcatee-azure: Looks good.

tompetersjr:
@igregson I talked to Casey and our work was somewhat overlapping so he will be taking on the order case, weight and volume update.  Can you make a PR that expands on this ones concept with the following idea:

```js
'order': {
    'computed': {
        'shippedCases': {
            'frozen': 1,
            'refrigeratedAndRoomTemp': 1,
            'greenhouse': 0,
            'total': 2
        },
        'shippedWeight': {
            'frozen': 1,
            'roomTemp': 0,
            'refrigerated': 1,
            'greenhouse': 1,
            'total': 3,
        },
        'orderedWeight': {
            'frozen': 1,
            'roomTemp': 0,
            'refrigerated': 1,
            'greenhouse': 1,
            'total': 3,
        },
        'shippedVolume': {
            'frozen': 1,
            'roomTemp': 0,
            'refrigerated': 1,
            'greenhouse': 1,
            'total': 3,
        },
        'orderedVolume': {
            'frozen': 1,
            'roomTemp': 0,
            'refrigerated': 1,
            'greenhouse': 1,
            'total': 3,
        }
    }
}
```

PR #245, opened 2018-11-14
Add inline options to order endpoints

igregson:
Adds `customer` to the inline options of the order endpoints. 

Questions:

- What is the preferred way of documenting this in the response object (more thoughts on this below)?
- We're sometimes returning inlined properties inside of an inline object (e.g. `inline: { customer: {...} }`) and other times not (e.g. there would be no `inline` property returned, customer would be at the top level in the return object). The order end points may be the only ones that currently return inside of an inline property. @davidmcatee-azure When discussing this while you were working on https://github.com/azurestandard/beehive/pull/4916 we settled on not nesting in an inline property. Thinking about the above question (how to document inlined (maybe returned) properties in the response), however, I'm questioning this. Having all maybe-returned properties nested in `inline` keeps the documentation cleaner. Here's an example (from ReDoc docs for this PR): 

image

Note: 

For the purposes of the Driver App, the only person properties needed are the person's id, name, and phone. In other places we've used a dot convention to specify specific sub properties... not sure what the best approach is here. Also, I've just realized that the person's phone is on the address endpoint (though it's not documented in the address response). Making matters worse, we have the awkward phone/text bit with the address... I suppose this means we need more inline options here on the order endpoints (`address.phone`, `address.text`). I'm going to go ahead and update this PR to include those, but let me know if this doesn't work for some reason and I'll change it as appropriate.

---

_I've updated the documentation for this PR a number of times (above). I'll keep the above for the backstory but have summarized the changes below._

Details: 

- Added `address` properties
- Improved the order GET endpoints and their response objects to better display in documentation tools
- ~~Added undocumented `phone` and `text` properties to the `address` response object~~


public.yaml
```diff
@@ -6656,6 +6639,25 @@ components:
...
+        type: array
+        items:
+          type: string
+          enum:
+            - "dropMembershipId"
+            - "customer.id"
+            - "customer.name"
+            - "customer.email"
+            - "address.phone"
+            - "address.text"
```
davidmcatee-azure, line 6660: The `address` associated with an `order` is actually used for the parcel carrier shipping destination. So this is not what we want. We want the phone and text of the customer associated with the order. But those _should_ come by default on the customer, so they don't need to be explicitly requested using the dot notation (see my comment below). Side note: `text` isn't yet a returned even though it should be (https://github.com/azurestandard/beehive/issues/4780).

Therefore the only ones needed here are `dropMembershipId` and `customer`. I don't think we need to list out the possible nested inlines here because they are the same inline properties that are supported by the person endpoint, e.g. `customer.is-drop-coordinator` would work here too (the same principal applies for all nested inlines, a point we should make clear somewhere but I'm not exactly sure where).

public.yaml
```diff
@@ -6656,6 +6639,25 @@ components:
       name: cookie
       description: |-
         Session cookie (http://tools.ietf.org/html/rfc6265).
+  parameters:
+    orderInline:
```
davidmcatee-azure, line 6643: I'm still not a fan of using the dedicated `inline` object despite it maybe resulting in cleaner documentation. It would result in `customer.inline.email`. No thanks. Also, I can imagine that someday we may decide to make `email` a permanent property on the `person` object, in which case client code would not need to change from `customer.inline.email` to `customer.email` because it would already be using `customer.email`.

public.yaml
```diff
@@ -9611,6 +9641,16 @@ components:
...
+        phone: 
+          description: |-
+            Phone number to use if there's no text number returned.
+          type: string
+          format: telephone
+        text: 
+          description: |-
+            Phone number where the user has indicated they may receive SMS messages. Prefer this number if returned.
+          type: string
+          format: telephone
```
davidmcatee-azure, line 9653: Given the comment above, these `address` fields are not relevant to this PR.

public.yaml
```diff
@@ -8219,6 +8221,34 @@ components:
                 The membership ID of the customer making the order. Only included if the request includes the inlined "dropMembershipId".
               type: integer
               format: int64
+            customer: 
+              description: |-
+                Each property is only returned if the request includes the lookup (e.g. `customer.id`).
```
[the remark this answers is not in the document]
igregson: This makes sense. Thanks for clarifying. It was an oversight on my part, not fully making sense of how the inline functionality works. I've edited as appropriate.

public.yaml
igregson: You're right. I've removed the `address` entries here and I agree that we shouldn't need to list of the possible nested inlines in these cases.

public.yaml
igregson: Good points. I'm fine with no dedicated `inline` object. 

I've updated this PR accordingly. Note that there _is_ currently a dedicated `inline` object for the order endpoint. From what I've been able to find, there's only one place we're using it (in the website's DC section, [here](https://github.com/azurestandard/website/blob/master/src/app.dc/dc_drop.order.pug#L27)). We'll need to update that once the API's response matches what's in this PR.

PR #247, opened 2018-11-30
Add orderFrequencySegment endpoint and computed property to person

igregson:
For https://github.com/azurestandard/beehive/issues/4119.

### DIFF
--- a/public.json
+++ b/public.json
@@ -7168,7 +7168,7 @@
         }
       }
     },
-    "/orderFrequencySegment": {
+    "/order-frequency-segment": {
       "get": {
         "summary": "Returns whether the customer has ordered in the last 120 days, \"active\" if yes or \"activeag\" (for \"active aging\") if not. If no customer exists, returns \"new\".",
         "operationId": "orderFrequencySegment",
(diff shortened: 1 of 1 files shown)


public.json
```diff
@@ -7167,6 +7167,37 @@
           }
         }
       }
+    },
+    "/orderFrequencySegment": {
```
caseybessette: We want endpoints to still use dashes, no?  The endpoint URL isn't JavaScript.

igregson: Hmm... doing some research on this it seems that hyphenated endpoints are more standard. Given that, I'll change this unless there's any reason we do want to go with camel cased endpoints.

public.json
```diff
@@ -7167,6 +7167,37 @@
...
+            "name": "email",
+            "in": "body",
+            "description": "Email of the person to look up.",
+            "required": true
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "Whether the customer has ordered in the last 120 days, \"active\" if yes or \"activeag\" (for \"active aging\") if not. If no customer exists, returns \"new\".",
+            "type": "string",
```
caseybessette: I would like to return an object here, e.g.,

```javascript
{
    'email': 'whatever was requested',
    'orderFrequencySegment': 'new'
}
```

PR #248, opened 2018-11-30
Fix case climates to match what we use

tompetersjr:


tompetersjr:
Upon further investigation I found that `cases-shipped` was already being used in a way that contradicted the specification.  Give that I left `cases-shipped` and fixed it to match the current implementation and added a new object of `casesShippedBreakdown`.

public.json
```diff
@@ -9054,23 +9059,18 @@
       "description": "Quantity of shipped boxes and bulk stock",
```
igregson: Seems we should add "by temperature type" to the description here. What do you think?

igregson: Actually, "quantity" doesn't seem to belong here at all. 

How about this:

> Breakdown of shipped boxes and bulk stock by temperature type.

public.json
```diff
@@ -9051,26 +9056,21 @@
       ]
     },
     "orderCases": {
```
igregson: The descriptions look good. 

The only of thing I see is that `orderCases` as a definition doesn't seem accurate. When I read "order cases" as a definition I expect a definition for an array of cases in an order... perhaps a better term to use for this definition would be `orderCaseQuantitiesByTemperature`?

Another thought is to remove the definition entirely.

PR #250, opened 2018-12-20
Update order-frequency-segment endpoint

igregson:
For response changes requested to the `order-frequency-segment` endpoint via https://github.com/azurestandard/website/issues/2582#issuecomment-448876545.

### DIFF
--- a/public.json
+++ b/public.json
@@ -7170,7 +7170,7 @@
     },
     "/order-frequency-segment": {
       "get": {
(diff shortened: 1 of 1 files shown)


caseybessette:
Why are you dropping the computed property?

igregson:
> Why are you dropping the computed property?


> Reasoning: This was not used in the implementation and since the spec has changed removing it seems more logically than updating it.

In other words: In the final implementation we decided against using it.

caseybessette:
Oops, I missed that commit message.

PR #255, opened 2019-02-06
Add product.maxRewardsRate.

caseybessette:
And update rewards-related descriptions to account for our change in
logic.

### DIFF
--- a/public.json
+++ b/public.json
@@ -8606,7 +8606,13 @@
           "minimum": 0
         },
         "rewards-rate": {
-          "description": "Rewards rate (as a percentage) earned for this packaging.  This will only be set for packaging with a fixed rewards rate for the given price level, regardless of the ordering customer.  When unset, the rewards-rate for this package is the customer's default rewards-rate.  The rewards earned for the packaging will be `dollars` * `rewards-rate` / 100, rounded to the nearest cent, rounding half to even.",
+          "description": "Rewards rate (as a percentage) earned for this packaging.  This will only be set for packaging with a fixed rewards rate.  The rewards earned for this product will be the price (after sales) * min(max(`person.rewards-rate`, `product.rewards-rate`), `product.maxRewardsRate`) / 100, rounded to the nearest cent, rounding half to even.",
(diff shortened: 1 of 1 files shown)


public.json
```diff
@@ -8606,7 +8606,13 @@
           "minimum": 0
         },
         "rewards-rate": {
-          "description": "Rewards rate (as a percentage) earned for this packaging.  This will only be set for packaging with a fixed rewards rate for the given price level, regardless of the ordering customer.  When unset, the rewards-rate for this package is the customer's default rewards-rate.  The rewards earned for the packaging will be `dollars` * `rewards-rate` / 100, rounded to the nearest cent, rounding half to even.",
+          "description": "Rewards rate (as a percentage) earned for this packaging.  This will only be set for packaging with a fixed rewards rate.  When unset, the rewards-rate for this package is the customer's default rewards-rate reduced by the maxRewardsRate for this product.  The rewards earned for the packaging will be `dollars` * `rewards-rate` / 100, rounded to the nearest cent, rounding half to even.",
```
davidmcatee-azure: I think this needs to mention the equation too `price (after sales) * min(max(`person.rewards-rate`, `product.rewards-rate`), `product.maxRewardsRate`) / 100`. Otherwise it sounds from reading this like the customer's rewards rate is not considered unless this is unset.

public.json
```diff
@@ -8606,7 +8606,13 @@
...
+          "type": "integer",
+          "format": "int32",
+          "minimum": 0
+        },
+        "maxRewardsRate": {
+          "description": "The maximum rewards rate that could be earned for this product.  A value of 0 means that no rewards will be earned.",
```
davidmcatee-azure: This can be unset, right?  And if it is not set, it is not used in the calculation?

davidmcatee-azure: Nevermind, I just looked at your code and saw that it defaults to 100. So if it is never unset, would that make this a `required` field?

caseybessette:
I've pushed a fixup commit addressing your suggestions.

PR #258, opened 2019-02-14
Changes for parcel out of stock handling

igregson:
For https://github.com/azurestandard/website/issues/2640.

### DIFF
--- a/public.json
+++ b/public.json
@@ -4419,6 +4419,25 @@
               "$ref": "#/definitions/changedOrder"
             }
           },
+          "409": {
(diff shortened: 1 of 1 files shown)


public.json
```diff
@@ -3206,6 +3206,18 @@
...
             "collectionFormat": "csv"
           },
+          {
+            "name": "code",
+            "in": "query",
+            "description": "packaged product codes to filter by",
+            "required": false,
+            "type": "array",
+            "items": {
+              "type": "integer",
```
davidmcatee-azure: Should be `string`

PR #260, opened 2019-04-06
Convert to OpenAPI 3.0 Spec

tompetersjr:
This moves us to the current spec from our current Swagger 2.0 json file.


tompetersjr:
Note that this should make us 100% valid with the specification.  Our old code could not validate as we were using 3.0 features in the 2.0 file.  In addition there were many syntax errors in the original file that are now fixed.

igregson:
Would it be preferable to first update the spec to 3.0 and then change the language to YAML (perhaps this would result in a cleaner Git history?).

Also (I suspect these items would best be handled via later PRs but wanted to at least mention them here):

- Do we want to alphabetize the endpoints and definitions?
- Do we want to standardize our spacing and punctuation? Some areas that currently are not consistent: 
  - Spacing between sentences (currently there's both single and double spaces used)
  - Capitalization (currently the first words of sentences are sometimes capitalized and sometimes not)
  - Periods at the end of sentences
- Do we want to update descriptions and summaries to use markdown where appropriate (I'm mostly thinking about inline code, namely the referencing of property names, but perhaps there are other places where markdown would improve these).
- Line breaks in property values (primarily in the description and summary fields) seem to all be around column 80... Without any type of formatter auto-handling this it feels like an arbitrary maintenance concern. Do we want to:
  1. Try to enforce the linebreaks somehow
  2. Take out the line breaks
  3. Other?

tompetersjr:
> Would it be preferable to first update the spec to 3.0 and then change the language to YAML (perhaps this would result in a cleaner Git history)?

I thought about that a good deal but the amount of cleanup needed to the JS code was significant just to make it compliant with 2.0.  And then there were items that could not be made compliant until 3.0.  The tool I used ([Swagger Hub](https://app.swaggerhub.com)) worked best with YAML.  In hind sight we probably should have all discussed before I went down this road. So while I agree with you I'd prefer not to spend the amount of time it would take to do that completely by hand. 

In the long run I feel that this data reads much better as YAML as well but that is personal preference.

As for the other comments I 100% agree with them and thought about several of them while doing this.  They can be done as separate PR's.

PR #263, opened 2019-05-29
Add id to route data

tompetersjr:


public.yaml
```diff
@@ -6948,6 +6948,8 @@ components:
       description: an order-delivery truck route
       type: object
       properties:
+        id:
+          type: integer
```
igregson: Looks like we need to add `format: int64` below this to be consistent with id property definitions elsewhere, no?

PR #273, opened 2020-08-03
Fix validation errors

igregson:
Found these inside of the Stoplight interface (https://azurestandard.stoplight.io/studio/api-spec/edit).

### DIFF
--- a/public.yaml
+++ b/public.yaml
@@ -1,7 +1,7 @@
 openapi: 3.0.0
 info:
-  version: 1.0.0-oas3
-  title: Azure Standard's Public API
+  version: "1.0.0-oas3"
+  title: Azure Standard Public API
   description: The public API behind Azure's website
   contact:
     name: Azure Standard Customer Service
@@ -16,7 +16,9 @@ security:
 paths:
   /affiliate-referrals:
     get:
-      summary: 'Returns all affiliate referrals for the given referring user, in descending order of the date the referred person''s account was created'
+      summary: >-
+        Returns all affiliate referrals for the given referring user, in
+        descending order of the date the referred person's account was created
       operationId: findAffiliateReferrals
       tags:
         - affiliate-referral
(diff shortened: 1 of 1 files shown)


public.yaml
```diff
@@ -16,9 +16,7 @@ security:
 paths:
   /affiliate-referrals:
     get:
-      summary: >-
-        Returns all affiliate referrals for the given referring user, in
-        descending order of the date the referred person's account was created
+      summary: 'Returns all affiliate referrals for the given referring user, in descending order of the date the referred person''s account was created'
```
davidmcatee-azure, line 19: Why are some description/summary lines enclosed in single quotes and others are not?

davidmcatee-azure: Yes, I agree to both. Could that be a separate PR?

igregson: Alright. Yes, I'll make the changes via a separate PR, along with a note about these guidelines, likely to the README.

igregson: I've reverted all of the auto-formatting that got applied via Stoplight. I'm still not sure how it was applied in the first place.

This leaves the summary and descriptions as they were (using YAML's `>-` for opening up a content block, which prevents the need to escape quotes and such...). I'd prefer to leave any changes here to a separate PR (it's formatting related, so perhaps in the same PR as the one to make casing and spacing consistent).

igregson:
Note that the `$ref` fixes here were to fix errors in the OpenAPI 3.0. The removal of the trailing slash was a warning. There are still a good number of other warnings (301, to be exact), but they're mostly due to missing fields, which seems best addressed via a separate PR.

image

public.yaml
igregson: > I'm still not sure how it was applied in the first place.

Learned that Stoplight does some autoformatting when editing via their UI. It removes all the `>-`, as well as line breaks in property values. Will need to address this via a separate formatting PR to prevent future conflicts/confusion over this, as well as make a note of it in a style spec.

PR #274, opened 2020-08-03
Improve formatting consistency

igregson:
Details: 

- Adds a "Style Guide" section to the readme.
- Makes formatting changes to match those specified in the style guide.
- Removes the block start syntax (`>-`) from `summary` properties (as `summary` should always be a simple string)
- Replaces the block start syntax (`>-`) on the `description` properties with `|-` (see https://yaml-multiline.info/ for differences). This is to support new lines in the description values, which are utilized by documentation tools. Also, when editing in Stoplight.io, this is the block start syntax that gets used, so another motivation for this is to prevent merge conflicts when editing between different tools.

### DIFF
--- a/public.yaml
+++ b/public.yaml
@@ -2,7 +2,8 @@ openapi: 3.0.0
 info:
   version: 1.0.0-oas3
   title: Azure Standard Public API
-  description: The public API behind Azure's website.
+  description: |-
+    The public API behind Azure's website.
   contact:
     name: Azure Standard Customer Service
     email: info@azurestandard.com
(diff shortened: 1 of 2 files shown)


public.yaml
```diff
@@ -527,42 +555,45 @@ paths:
...
             application/json:
               schema:
                 $ref: '#/components/schemas/error'
   '/bartender-print-configuration/{id}/actions/test-print':
     post:
-      summary: >-
-        Create a test BarTender print request using this print configuration.
-        The print request uses hard-coded dummy data if a stock ID is not
-        provided.
+      summary: Create a test BarTender print request using this print configuration. The print request uses hard-coded dummy data if a stock ID is not provided.
```
davidmcatee-azure, line 577: This summary is formatted like a description. Maybe it should be changed to be a `description`. A summary could be added such as "Create a test BarTender print request"

public.yaml
```diff
@@ -1254,40 +1309,44 @@ paths:
...
             application/json:
               schema:
                 $ref: '#/components/schemas/error'
   /drops/locations:
     get:
-      summary: >-
-        Returns an array of locations for all active drops that the user has
-        access to.
-      description: 'For example, you can use this to draw markers on map.'
+      summary: Returns an array of locations for all active drops that the user has access to.
```
davidmcatee-azure, line 1331: Another summary ending in a period.

public.yaml
```diff
@@ -4339,37 +4521,40 @@ paths:
...
             application/json:
               schema:
                 $ref: '#/components/schemas/error'
   /account-entries:
     get:
-      summary: >-
-        Returns all dollar credits and debits from the system that the user has
-        access to.  Results are ordered by increasing settled dates, falling
-        back to IDs when multiple entries settled on the same day.
+      summary: Returns all dollar credits and debits from the system that the user has access to.
```
davidmcatee-azure, line 4547: Another summary ending in a period.

public.yaml
```diff
@@ -4628,33 +4831,36 @@ paths:
...
             application/json:
               schema:
                 $ref: '#/components/schemas/error'
   /rewards:
     get:
-      summary: >-
-        Returns all reward credits and debits from the system that the user has
-        access to.  Results are ordered by increasing settled times, falling
-        back to ID for pending debits.
+      summary: Returns all reward credits and debits from the system that the user has access to.
```
davidmcatee-azure, line 4853: Another summary ending in a period.

public.yaml
igregson: Good catch. There are lots of other summaries with a bit too much information, where that extra information could be moved to descriptions. I held off on doing too much of this in this PR since it seems best suited to its own scope.

PR #276, opened 2020-08-04
Add inline customer to order endpoints

igregson:


igregson:
Seems Stoplight's "Studio" is still doing things it shouldn't to our spec file :(

Opened an issue for it: https://github.com/stoplightio/studio/issues/484

Will fix here (or create a new PR).

igregson:
Accidentally closed this. Then learned that this was a duplicate of #245. Will leave this closed in favor of that one.

PR #277, opened 2020-08-21
Wire up Redoc

igregson:


igregson:
Wasn't sure who to have review this. Since it deals with our team's workflow having you look at it, @tompetersjr, seemed good; since it includes a Node.js single-file (zero dependency) HTTP server it seemed good to have you look at it, @davidmcatee-azure.

igregson:
Regarding `serve.js`: 

My thinking was that it's useful to be able to quickly view the docs site locally. Thinking this through a bit more, however, it seems that the best usecase for this is if/when editing the spec site directly (locally), rather than going through a GUI (such as Stoplight.io). In this case, we'd want a watch command that will auto-reload on change of the spec file. If going this route, we'll need to include a few dependencies, which would allow us to further simplify the serve handling. I can do that in the context of this PR or in a later follow-up one, whichever seems preferable.

igregson:
I've just learned about the `redoc-cli`, which has a `bundle` command. This generates a zero-dependency html file that loads much faster than the current version (loading Redoc from a CDN). 

To test this locally: 

```sh
$ npx redoc-cli bundle public.yaml
```

Then:

```sh
$ node serve.js
```

And go to http://localhost:8080/redoc-static.html


This may be a preferable approach, especially since the site is going to be much more read much more than built/compiled. What do you think, @davidmcatee-azure?

Issue #289, opened 2020-09-30
Handle Truck Number in API


tompetersjr:
# Background
Trips can have multiple trucks (PickSets).  At this point the API has no way of knowing what truck is delivering for a stop.  The DriverApp could use this info so that a driver is only working with their stops and not a list of stops that could be delivered on other trucks.  In addition the logistics application could also benefit from the ability to group by truck as well

**Note:** While we are not adding driver info specific to each truck that is something we want to consider for the future.  Right now the driver info is stored at the route level.  It should really be at the pickset level.

### Trip Endpoint

- On the trip endpoint add a `truckIds` property as an array.  The array would be PickSet id's that are available in the given trip. 
- This array would be undefined if the trip is not yet verified as trucks are not setup until verification time.

### Stop Endpoint

- Allow the stop endpoint to be filtered by `truckId`.
- On the stop endpoint add a `truckId` and `truckNumber` property as integers.
- These properties would be undefined if the trip is not yet verified.

### Trip Endpoint

- On the trip endpoint add a `truckIds` property as an array.  The array would be PickSet id's that are available in the given trip. 
- This array would be undefined if the trip is not yet verified as trucks are not setup until verification time.

### Stop Endpoint

- Allow the stop endpoint to be filtered by `truckId`.
- On the stop endpoint add a `truck` object.  Within that object have a property for `id` and `number` as integers.  Then as we add `driverName` and `driverPhone` in the future it would also appear in this object.
- The `truck` object would be undefined if the trip is not yet verified.

davidmcatee-azure:
If we were to make it similar to how the rest of the API is structured, trucks (PickSets) would have their own endpoint.

While the name "truck" is more user-friendly, when you think of it as an entity, it sounds like it would represent a specific physical truck.

So maybe something like "truck-load" would be more accurate. But I think we could actually keep the name "pick-set" to avoid confusion, since this is something only we would use.

I'm thinking their would be an API endpoint for pick-sets (with singular and plural). The object returned from it would have `id` and `truckNumber` properties (and could include more such as total weight and volume for example).  In the future we could add the `driverId` (not `driverName` and `driverPhone`, because the driver would be retrieved from the `person` endpoint using the `driverId`, or it could be inlined onto the `pick-set` object using `?inline=driver` as we have done elsewhere).

So in summary:

### Trip Endpoint

- No changes required. PickSets for a given trip could be retrieved using `/pick-sets?tripId={tripId}`.

### Stop Endpoint

[...]

PR #291, opened 2020-10-01
API Truckload Support

tompetersjr:
Add support for a truckload concept to the API.

Fixes: #289

### DIFF
--- a/spec.json
+++ b/spec.json
@@ -9904,9 +9904,16 @@
           },
           "truckloadIds": {
             "type": "array",
-            "description": "Array of truck  ID's. ",
+            "description": "Array of truckload ID's. ",
             "items": {
               "type": "integer"
+            }
+          },
+          "truckloads": {
+            "type": "array",
+            "description": "Array of truckloads. This will only be included if the `inline` query parameter includes `truckloads`.",
+            "items": {
+              "$ref": "#/components/schemas/truckload"
             }
           }
         },
@@ -10056,7 +10063,7 @@
           },
           "truckloadId": {
             "type": "integer",
-            "description": "ID of the truck (pickset)."
+            "description": "ID of the truckload (pickset)."
           }
         },
         "required": [
@@ -12830,6 +12837,25 @@
             "type": "string"
           }
         }
+      },
+      "truckload": {
+        "title": "truckload",
+        "type": "object",
+        "description": "A truck making stops on a trip. There can be multiple truckloads per trip. There is no truckload associated with a trip until it is verified.",
+        "properties": {
+          "id": {
+            "type": "integer",
+            "description": "ID of the truckload (pickset)."
+          },
+          "truckNumber": {
+            "type": "integer",
+            "description": "Number of the truck on the trip. This is not necessarily sequential."
+          }
+        },
+        "required": [
+          "id",
+          "truckNumber"
+        ]
       }
     }
   }
--- a/spec.json
+++ b/spec.json
@@ -10085,7 +10085,7 @@
           },
           "truckloadId": {
             "type": "integer",
-            "description": "ID of the truckload (pickset). This is only included if the trip is verified."
+            "description": "ID of the truckload (pickset). This will only be included if the inline query parameter includes truckloadId and the trip is verified."
           }
         },
         "required": [
--- a/spec.json
+++ b/spec.json
@@ -2845,7 +2845,7 @@
             "schema": {},
             "in": "query",
             "name": "inline",
-            "description": "For convenience, include additional information. Currently supports \"truckloadIds\" as described in the trip model."
+            "description": "For convenience, include additional information. Currently supports \"truckloads\" as described in the trip model."
           }
         ],
         "responses": {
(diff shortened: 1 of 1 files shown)


spec.json
```diff
@@ -9901,13 +9901,28 @@
               "sequenced",
               "shipped"
             ]
+          },
+          "truckloadIds": {
+            "type": "array",
+            "description": "Array of truckload ID's. ",
```
igregson, line 9907: Two things here: 
1. Since we've done so elsewhere, we should probably include "(pickset)" after "truckload"
2. There's a trailling space

spec.json
```diff
@@ -9901,13 +9901,28 @@
...
+            "items": {
+              "type": "integer"
+            }
+          },
+          "truckloads": {
+            "type": "array",
+            "description": "Array of truckloads. This will only be included if the `inline` query parameter includes `truckloads`.",
```
igregson, line 9914: Since we've done so elsewhere, we should probably include "(picksets)" after "truckloads"

igregson (changes requested):
Noticed two small description-related things.

igregson (approved):

spec.json
```diff
@@ -9901,13 +9901,28 @@
...
+            }
           }
         },
         "required": [
           "id",
           "route",
           "cutoff",
-          "delivery-start"
+          "delivery-start",
+          "truckloadIds"
```
davidmcatee-azure, line 9925: Since `truckloadIds` will be not be present if the trip is not verified, I don't think it should be required. We may also want to mention this in the description for `truckLoadIds` above.

spec.json
```diff
@@ -10045,14 +10060,19 @@
...
           }
         },
         "required": [
           "id",
           "trip",
           "estimatedDelivery",
           "finalizedDelivery",
-          "actualDeliveryBackdated"
+          "actualDeliveryBackdated",
+          "truckloadId"
```
davidmcatee-azure, line 10075: Since `truckloadId` will be not be present if the stop's trip is not verified, I don't think it should be required. We may also want to mention this in the description for `truckLoadId` above.

davidmcatee-azure (changes requested):

spec.json
igregson: I had understood that it would still be included, but set to `undefined`, no? Same for your point below.

davidmcatee-azure: That would be breaking from the API's existing pattern of not including such properties.

davidmcatee-azure (approved):

spec.json
```diff
@@ -2840,6 +2840,12 @@
               "type": "integer",
               "format": "int32"
             }
+          },
+          {
+            "schema": {},
+            "in": "query",
+            "name": "inline",
+            "description": "For convenience, include additional information. Currently supports \"truckloadIds\" as described in the trip model."
```
davidmcatee-azure, line 2848: Should be `truckloads` here, not `truckloadIds`.

spec.json
```diff
@@ -2936,6 +2942,14 @@
...
               "format": "int64"
             }
+          },
+          {
+            "schema": {
+              "type": "array"
+            },
+            "in": "query",
+            "name": "inline",
+            "description": "For convenience, include additional information. Currently supports \"truckloadIds\" as described in the trip model."
```
davidmcatee-azure, line 2952: Should be `truckloads` here, not `truckloadIds`.

spec.json
```diff
@@ -10045,6 +10082,10 @@
           "actualDeliveryBackdated": {
             "type": "boolean",
             "description": "The acutal delivery was manually set to a datetime in the past."
+          },
+          "truckloadId": {
+            "type": "integer",
+            "description": "ID of the truckload (pickset). This is only included if the trip is verified."
```
davidmcatee-azure, line 10088: "This will only be included if the `inline` query parameter includes `truckloadId` and the trip is verified."

igregson: @tompetersjr Other than this, your changes look good to me.

tompetersjr: To make that statement true I need to make a change to the beehive PR [here](https://github.com/azurestandard/beehive/pull/5048/files#diff-61480bfff576892bbced73b51d655f951e9bd1131930996a06f53fe6d510f44dR40-R44).  I'll make that change as well.

spec.json
```diff
@@ -10045,6 +10082,10 @@
           "actualDeliveryBackdated": {
             "type": "boolean",
             "description": "The acutal delivery was manually set to a datetime in the past."
+          },
+          "truckloadId": {
+            "type": "integer",
+            "description": "ID of the truckload (pickset). This will only be included if the inline query parameter includes truckloadId and the trip is verified."
```
igregson, line 10088: Based on formatting elsewhere in the spec, "truckloadId" here should either be in double quotes or backticks (it seems we're not consistent with which of these we prefer, so either would be fine for now, IMO).

igregson (approved):

igregson:
This had merge conflicts due to #293. For some reason the way I was resolving them kept breaking things but I think it's now (finally) right. @tompetersjr / @davidmcatee-azure want to give this another quick look before we merge it?

tompetersjr:
Looks good to me.

davidmcatee-azure (approved):

PR #292, opened 2020-10-15
Add Development Site Link

tompetersjr:


README.md
```diff
@@ -10,4 +10,8 @@ OpenAPI 3.0 Specification for Azure Standard's API.
 
 URL: https://azure-api-docs.netlify.app/
 
-Uses [Redoc](https://github.com/Redocly/redoc).
\ No newline at end of file
+Uses [Redoc](https://github.com/Redocly/redoc).
+
+## Development site
```
igregson, line 15: For some reason "Development site" for this doesn't sound quite right to me. Something like "Editing UI" seems more accurate, no?

Another approach would be to use "Stoplight" or "Stoplight Studio" for the heading. Then, under the heading, we could briefly touch on what stoplight is / why we use it, and include a link to our instance/account there. Something like: 

---

## Stoplight Studio

We have an account at [stoplight.io](https://stoplight.io/), which includes their "Studio" (a visual OpenAPI editor).

Stoplight account: https://azurestandard.stoplight.io/
Stoplight Studio: https://azurestandard.stoplight.io/studio/api-spec/edit

---

I'm fine with either approach. From an onboarding/zero-past-knowledge perspective, the more verbose seems preferable.

README.md
```diff
@@ -10,4 +10,11 @@ OpenAPI 3.0 Specification for Azure Standard's API.
 
 URL: https://azure-api-docs.netlify.app/
 
-Uses [Redoc](https://github.com/Redocly/redoc).
\ No newline at end of file
+Uses [Redoc](https://github.com/Redocly/redoc).
+
+## Stoplight Studio
+
+We have an account at stoplight.io, which includes their "Studio" (a visual OpenAPI editor).
```
igregson, line 17: Looks like GitHubs markdown isn't auto linking this, so we'll need to explicitly make it a link (`stoplight.io` -> `[stoplight.io](https://stoplight.io/)`).

README.md
```diff
@@ -10,4 +10,11 @@ OpenAPI 3.0 Specification for Azure Standard's API.
...
+
+Stoplight account: https://azurestandard.stoplight.io/
+Stoplight Studio: https://azurestandard.stoplight.io/studio/api-spec/edit
```
igregson, line 20: This is rendering as a single paragraph ([see here](https://github.com/azurestandard/api-spec/tree/tjp/dev-site-readme#stoplight-studio)). A double space at the end of the first line should fix it.

README.md
```diff
@@ -10,4 +10,11 @@ OpenAPI 3.0 Specification for Azure Standard's API.
 
 URL: https://azure-api-docs.netlify.app/
 
-Uses [Redoc](https://github.com/Redocly/redoc).
\ No newline at end of file
+Uses [Redoc](https://github.com/Redocly/redoc).
+
+## Stoplight Studio
+
+We have an account at [stoplight.io](https://stoplight.io/) which includes their "Studio" (a visual OpenAPI editor).
```
igregson, line 17: I'm pretty sure there should be a comma between the stoplight.io link and "which". I.e. 

```diff
-We have an account at [stoplight.io](https://stoplight.io/) which includes their "Studio" (a visual OpenAPI editor).
+We have an account at [stoplight.io](https://stoplight.io/), which includes their "Studio" (a visual OpenAPI editor).
```

igregson: Looks like this was there but got removed via b7733228e62b1d7dfc6ce84f9919aeca55f017be.

igregson (approved):

PR #293, opened 2020-10-19
Update stop endpoints and response

### LINKED ISSUE #281: [API parity] Document inline=drop,pickup, dropId and pickupId on stop endpoints
Landed in the API via https://github.com/azurestandard/beehive/pull/4916.

Need to document:

- `drop` (drop object, returned only if `inline=drop` is passed and the stop is a drop)
- `pickup` (pickup object, returned only if `inline=pickup` is passed and the stop is a pickup)
- `dropId` (`int`) (always returned if stop is a drop)
- `pickupId` (`int`) (always returned if stop is a pickup)

igregson:
Resolves #281

### DIFF
--- a/spec.json
+++ b/spec.json
@@ -10101,8 +10101,7 @@
           "estimatedDelivery",
           "finalizedDelivery",
           "actualDeliveryBackdated"
-        ],
-        "title": ""
+        ]
       },
       "updateStop": {
         "description": "An update to a trip stop or waypoint.",
--- a/spec.json
+++ b/spec.json
@@ -3160,6 +3160,9 @@
               "type": "integer",
               "format": "int32"
             }
+          },
+          {
+            "$ref": "#/components/parameters/stopInline"
           }
         ],
         "responses": {
@@ -3267,6 +3270,9 @@
               "type": "integer",
               "format": "int64"
             }
+          },
+          {
+            "$ref": "#/components/parameters/stopInline"
           }
         ],
         "responses": {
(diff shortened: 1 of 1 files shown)


spec.json
```diff
@@ -9015,6 +9021,23 @@
...
         }
+      },
+      "stopInline": {
+        "name": "inline",
+        "in": "query",
+        "description": "Optional properties returned if specified in the request.",
+        "required": false,
+        "style": "form",
+        "schema": {
+          "type": "array",
```
davidmcatee-azure, line 9032: `inline` is not really an array of strings, but a string that is a comma-separated list.

davidmcatee-azure (changes requested):

spec.json
igregson: Argh... just kidding. It seems Redoc doesn't like the change I made. Looking into this further...

igregson: Alright, if we specify this as a string and use `enum` for the possible values, here's how Redoc shows it: 

image

I'm not 100% sure this is what we want, as it's not clear that the string is comma-separated. If we use `array` as the type, isn't this more clear? Here's how Redoc show it is using `array`: 

image

davidmcatee-azure: I'd be OK keeping it as an array of strings, as long as the comma-separated aspect is made clear somewhere.

igregson: As for how we want to make the comma-separated aspect clear, it seems that Redoc doesn't yet support example requests (only example responses) ([see open issue here](https://github.com/Redocly/redoc/issues/15)). 

I'm thinking the simplest thing to do for now is to specify this in the top-level `description`. That seems best via a different PR, yes?

davidmcatee-azure: Sounds good.

davidmcatee-azure (approved):

PR #294, opened 2020-10-20
Add mail/route and mail/trip endpoints

### LINKED ISSUE #284: [API parity] Document mail endpoints (route and trip)
This endpoint is not currently in the spec and, as far as I know, never was.

It's being used [here](https://github.com/azurestandard/internal-website/blob/master/src/modules/recall/recall.js) and [here](https://github.com/azurestandard/internal-website/blob/master/src/modules/logistics/controllers/modal.sendMessage.controller.js#L83) in the Logistics app.

igregson:
Resolves #284 

Also fixes two misspellings (unrelated to the issue this resolves).

### DIFF
--- a/spec.json
+++ b/spec.json
@@ -9117,7 +9117,7 @@
         },
         "in": "query",
         "name": "message",
-        "description": "The conent of the email."
+        "description": "The content of the email."
       },
       "mailSubject": {
         "schema": {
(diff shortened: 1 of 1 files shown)


spec.json
```diff
@@ -9053,6 +9132,52 @@
...
+        "name": "subject",
+        "description": "The email's subject."
+      },
+      "mailTarget": {
+        "schema": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
+        },
```
davidmcatee-azure, line 9177: It looks like `target` should just be a `string`, not an array.

davidmcatee-azure (changes requested):

spec.json
igregson: Good catch. Fixed.

davidmcatee-azure (approved):

PR #295, opened 2020-10-20
Add drops/actions/convert-route endpoint

### LINKED ISSUE #283: [API parity] Document /drops/actions/convert-route
Landed in the API via https://github.com/azurestandard/beehive/pull/4738.

This is for moving a drop from one route to another.

It's being used [here in the Logistics app](https://github.com/azurestandard/internal-website/blob/master/src/modules/logistics/controllers/logistics.js#L205-L216).

igregson:
Resolves #283

### DIFF
--- a/spec.json
+++ b/spec.json
@@ -8977,7 +8977,7 @@
         "parameters": [
           {
             "schema": {
-              "type": "string"
+              "type": "integer"
             },
             "in": "query",
             "name": "sourceRouteId",
@@ -8985,7 +8985,7 @@
           },
           {
             "schema": {
-              "type": "string"
+              "type": "integer"
             },
             "in": "query",
             "name": "targetRouteId",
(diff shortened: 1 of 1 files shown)


igregson:
Wasn't sure who to request a review from for this. Went with you, David, since you've generally reviewed my previous API-Spec PRs. Since you worked on this code in the backend, Tom, thought maybe that your review on it would also/instead be appropriate.

spec.json
```diff
@@ -8964,6 +8964,44 @@
...
+            "name": "targetRouteId",
+            "description": "The ID of the route that the drop(s) should be moved to."
+          },
+          {
+            "schema": {
+              "type": "string"
+            },
+            "in": "query",
+            "name": "dropIds",
+            "description": "A comma-separated list of the IDs of drops to be moved."
```
davidmcatee-azure, line 9000: According to the code, `dropIds` is not a comma-separated list of ids, but rather an array of ids.

davidmcatee-azure (changes requested):

spec.json
```diff
@@ -8964,6 +8964,47 @@
...
+          "200": {
+            "description": "OK"
+          }
+        },
+        "parameters": [
+          {
+            "schema": {
+              "type": "integer"
+            },
+            "in": "query",
```
davidmcatee-azure, line 8982: Just noticed that these parameters should be in the request body, not the query parameters.

davidmcatee-azure (changes requested):

spec.json
igregson: 🤦 

Good catch. Fixed.

davidmcatee-azure (approved):

PR #297, opened 2020-10-22
Add timezone to drop and pickup response objects

igregson:
Related Beehive issue: https://github.com/azurestandard/beehive/issues/5065


igregson:
This landed in the API via https://github.com/azurestandard/beehive/pull/5073.

PR #302, opened 2020-12-04
WIP: Modified spec.json for Worldship

tompetersjr:
API documentation for Worldship application.

NOTE: UPS is working with this API as it is live.  However there are a few items I will clean up once they deliver their code.  These should have been addressed before hand but it is too late for me to change the spec on the UPS developers.

1. Use camel case for JSON body and response
2. The /worldship endpoint could have both an GET and PUT instead of creating an separate /worldship/import endpoint for PUT requests
3. The PUT request should use an array of strings for the tracking_code instead of a string separated by spaces.

### DIFF
--- a/spec.json
+++ b/spec.json
@@ -9105,6 +9105,87 @@
           }
         ]
       }
+    },
+    "/worldship/{order_id}": {
+      "get": {
+        "summary": "Your GET endpoint",
+        "tags": [],
+        "responses": {
+          "200": {
+            "description": "OK",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "description": "An array of shippments for the order. Azure allows customers to split there order between dry and frozen items.",
+                  "minItems": 1,
+                  "uniqueItems": true,
+                  "type": "array",
+                  "items": {
+                    "$ref": "#/components/schemas/worldship"
+                  }
+                }
+              }
+            }
+          }
+        },
+        "operationId": "get-worldship",
+        "requestBody": {
+          "content": {
+            "application/json": {
+              "schema": {
+                "type": "object",
+                "properties": {}
+              }
+            }
+          }
+        },
+        "description": "Get all the shippments for a specific order for Worldship to process and ship. Azure allows customers to split there order between dry and frozen items. As such this endpoint will return an array of shippments."
+      },
+      "parameters": []
+    },
+    "/worldship/import/{order_shipment_id}": {
+      "get": {
+        "summary": "Your GET endpoint",
+        "tags": [],
+        "responses": {},
+        "operationId": "get-worldship-import"
+      },
+      "parameters": [],
+      "put": {
+        "summary": "",
+        "operationId": "put-worldship-import-order_shipment_id",
+        "responses": {
+          "201": {
+            "description": "Created"
+          }
+        },
+        "requestBody": {
+          "content": {
+            "application/json": {
+              "schema": {
+                "description": "",
+                "type": "object",
+                "properties": {
+                  "actual_amount": {
+                    "type": "number",
+                    "description": "Amount charged via Worldship for the UPS shipping amount."
+                  },
+                  "tracking_code": {
+                    "type": "string",
+                    "minLength": 1,
+                    "description": "List of tracking codes sepparated by spaces."
+                  }
+                },
+                "required": [
+                  "actual_amount",
+                  "tracking_code"
+                ]
+              }
+            }
+          }
+        },
+        "description": "Import data from WorldShip into Beehive."
+      }
     }
   },
   "servers": [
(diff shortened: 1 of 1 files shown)


spec.json
```diff
@@ -9105,6 +9105,76 @@
...
+        "tags": [
+          "worldship"
+        ],
+        "responses": {
+          "200": {
+            "description": "OK",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "description": "An array of shippments for the order. Azure allows customers to split there order between dry and frozen items.",
```
igregson, line 9121: > there order

This should be the possessive form (i.e. "their").

However, I think it being plural reads better here (i.e. "to split orders"), making this sentence read:

> Azure allows customers to split orders between dry and frozen items.

spec.json
```diff
@@ -9105,6 +9105,76 @@
...
+                  "items": {
+                    "$ref": "#/components/schemas/worldship"
+                  }
+                }
+              }
+            }
+          }
+        },
+        "operationId": "get-worldship",
+        "description": "Get all the shippments for a specific order for Worldship to process and ship. Azure allows customers to split there order between dry and frozen items. As such this endpoint will return an array of shippments."
```
igregson, line 9134: > there order

Same as above.

spec.json
```diff
@@ -13084,7 +13154,114 @@
...
+          "address_2"
+        ]
+      },
+      "boxWeights": {
+        "title": "boxWeights",
+        "type": "array",
+        "x-examples": {},
+        "description": "Array of numbers for multiple box weights",
+        "items": {
+          "type": "number"
```
igregson, line 13257: This is missing `format` (pretty sure it's a `float`).

```json
"format": "float"
```

spec.json
```diff
@@ -13084,7 +13154,114 @@
...
         ]
+      },
+      "worldship": {
+        "description": "Data for the UPS Worldship application.",
+        "type": "object",
+        "properties": {
+          "profile": {
+            "type": "string",
+            "minLength": 1,
+            "description": "Worldship profile name"
```
igregson, line 13165: Following our [style guide](https://github.com/azurestandard/api-spec#style-guide) (that states that descriptions should use sentence formatting), this should have a period. Same goes for many of the descriptions below.

spec.json
```diff
@@ -13084,7 +13154,114 @@
...
+          },
+          "order_id": {
+            "type": "number",
+            "description": "Beehive order id"
+          },
+          "reason": {
+            "type": "string",
+            "minLength": 1,
+            "description": "Reason for split shippment.  Can be one of the following choices.\n\n- DRY\n- FROZEN\n- ALL"
```
igregson, line 13174: - Should use a single space between sentences.
- Rather than a list here, the pattern we've followed elsewhere in the spec for displaying possible return values is to use `enum`. Here's an example: https://github.com/azurestandard/api-spec/blob/master/spec.json#L11340-L11348. I'd suggest we follow that pattern here. If there's a reason not to (and to otherwise keep the list), the preceding sentence should end with a colon (`:`), not a period.

spec.json
```diff
@@ -13084,7 +13154,114 @@
...
+          },
+          "shipping_service": {
+            "type": "string",
+            "minLength": 1,
+            "description": "UPS shipping service for Worldship to use.  Can be one of the following choices.\n\n- UPS Ground\n- UPS 2nd Day Air"
```
igregson, line 13179: Same two points above apply here.

spec.json
```diff
@@ -13084,7 +13154,114 @@
...
+            "minLength": 1,
+            "description": "State to ship to"
+          },
+          "bill_to": {
+            "type": "string",
+            "minLength": 1,
+            "description": "Account to bill to in Worldship"
+          },
+          "order_shipment_id": {
+            "type": "number",
```
igregson, line 13220: This is missing `format`. Also, should it be an `integer`?

```json
"type": "integer",
"format": "int64"
```

spec.json
```diff
@@ -13084,7 +13154,114 @@
...
+        "description": "Data for the UPS Worldship application.",
+        "type": "object",
+        "properties": {
+          "profile": {
+            "type": "string",
+            "minLength": 1,
+            "description": "Worldship profile name"
+          },
+          "order_id": {
+            "type": "number",
```
igregson, line 13168: This is missing `format`. Also, should it be an `integer`?

```json
"type": "integer",
"format": "int64"
```

spec.json
```diff
@@ -9105,6 +9105,76 @@
...
+        },
+        "requestBody": {
+          "content": {
+            "application/json": {
+              "schema": {
+                "description": "",
+                "type": "object",
+                "properties": {
+                  "actual_amount": {
+                    "type": "number",
```
igregson, line 9156: This is missing `format` (pretty sure it's a `float`).

```json
"format": "float"
```

igregson (changes requested):
Other than some formatting bits (to align with [our style guide](https://github.com/azurestandard/api-spec#style-guide)), I found the `format` property missing in places and a type issue.

spec.json
```diff
@@ -9105,6 +9105,77 @@
...
+        "tags": [
+          "worldship"
+        ],
+        "responses": {
+          "200": {
+            "description": "OK",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "description": "An array of shippments for the order. Azure allows customers to split orders between dry and frozen items.",
```
igregson, line 9121: Misspelling: shippments -> shipments

This same misspelling occurs in other places, too, that should be fixed, but I won't comment on them.

spec.json
```diff
@@ -13084,7 +13155,126 @@
...
+            "minLength": 1,
+            "description": "Zipcode to ship to."
+          },
+          "box_weights": {
+            "$ref": "#/components/schemas/boxWeights"
+          },
+          "country": {
+            "type": "string",
+            "minLength": 1,
+            "description": "Country to ship to"
```
igregson, line 13218: This should also have a period.

spec.json
```diff
@@ -13084,7 +13155,126 @@
...
+            "description": "State to ship to."
+          },
+          "bill_to": {
+            "type": "string",
+            "minLength": 1,
+            "description": "Account to bill to in Worldship."
+          },
+          "order_shipment_id": {
+            "type": "integer",
+            "description": "The id of the shippment this applies to in. Azure allows customers to split orders between dry and frozen items.",
```
igregson, line 13232: > The id of the shippment this applies to in.

Delete "in" ?

Also, I think it would clearer if we said: 

> The id of the shipment this applies to (needed because Azure allows customers to split an order into multiple shipments by climate type).

spec.json
```diff
@@ -13084,7 +13155,126 @@
...
+          },
+          "address_1": {
+            "type": "string",
+            "minLength": 1,
+            "description": "Address to ship to."
+          },
+          "address_2": {
+            "type": "string",
+            "description": "Address to ship to."
+          }
```
igregson, line 13243: It seems that one should be able to understand the difference between these two (`address_1` and `address_2`) by reading the descriptions, no?

spec.json
```diff
@@ -9105,6 +9105,77 @@
...
+                  "items": {
+                    "$ref": "#/components/schemas/worldship"
+                  }
+                }
+              }
+            }
+          }
+        },
+        "operationId": "get-worldship",
+        "description": "Get all the shippments for a specific order for Worldship to process and ship. Azure allows customers to split orders between dry and frozen items. As such this endpoint will return an array of shippments."
```
igregson, line 9134: I think this would be more clear if it said:

> Gets all the shipments for a specific order for Worldship to process and ship. Note: An Azure order can be split into multiple shipments (by climate type), so this endpoint always returns an array of shipments (rather than a single shipment).

spec.json
```diff
@@ -9105,6 +9105,77 @@
...
+                "properties": {
+                  "actual_amount": {
+                    "type": "number",
+                    "description": "Amount charged via Worldship for the UPS shipping amount.",
+                    "format": "float"
+                  },
+                  "tracking_code": {
+                    "type": "string",
+                    "minLength": 1,
+                    "description": "List of tracking codes sepparated by spaces."
```
igregson, line 9163: Misspelling: sepparated -> separated

igregson (changes requested):

igregson:
@tompetersjr 

> NOTE: UPS is working with this API as it is live. However there are a few items I will clean up once they deliver their code. These should have been addressed before hand but it is too late for me to change the spec on the UPS developers.
>[...] 
> 3. The PUT request should use an array of strings for the tracking_code instead of a string separated by spaces.

With the above in mind, it seems `tracking_code` (once made camel case) should be plural (i.e. `trackingCodes`), since it'll be an array.

spec.json
```diff
@@ -13084,7 +13155,125 @@
...
+          "customer",
+          "city",
+          "shipper",
+          "zip",
+          "box_weights",
+          "state",
+          "bill_to",
+          "order_shipment_id",
+          "address_1",
+          "address_2"
```
igregson, line 13259: Shouldn't `address_2` be optional (not required)?

tompetersjr: Yes it should be.

igregson (approved):

PR #304, opened 2020-12-21
Add parcelCarrierShipByDate field to the spec for updateOrder endpoint

dwmorrison33:
This is for the latest code in PR: https://github.com/azurestandard/beehive/pull/5188 

@igregson I could not remember how to update this spec in the UI of https://azurestandard.stoplight.io/docs/api-spec/public.yaml so I just modified the code manually in the spec.json to add the field that is now getting returned in the `updateOrder` endpoint.

### DIFF
--- a/spec.json
+++ b/spec.json
@@ -11377,6 +11377,11 @@
           "holdShipmentUntilStockAvailable": {
             "description": "If true, the customer has requested to hold shipment of the order until all requested stock is available.",
             "type": "boolean"
+          },
+          "parcelCarrierShipByDate": {
+            "description": "For parcel carrier orders, this is the date the order needs to ship by.",
+            "type": "string",
+            "format": "date"
           }
         },
         "required": [
--- a/spec.json
+++ b/spec.json
@@ -11096,7 +11096,7 @@
             "type": "boolean"
           },
           "parcelCarrierShipByDate": {
-            "description": "For parcel carrier orders, this is the date the order needs to ship by.",
+            "description": "For parcel carrier orders, this is the date the order needs to ship by. Since we are using ISO format, the string can either be a date or datetime but it will be stored on the backend as a date.",
             "type": "string",
             "format": "date"
           },
@@ -11384,7 +11384,7 @@
             "type": "boolean"
           },
           "parcelCarrierShipByDate": {
-            "description": "For parcel carrier orders, this is the date the order needs to ship by.",
+            "description": "For parcel carrier orders, this is the date the order needs to ship by. Since we are using ISO format, the string can either be a date or datetime but it will be stored on the backend as a date.",
             "type": "string",
             "format": "date"
           }
--- a/spec.json
+++ b/spec.json
@@ -11095,6 +11095,11 @@
             "description": "If true, the customer has requested to hold shipment of the order until all requested stock is available.",
             "type": "boolean"
           },
+          "parcelCarrierShipByDate": {
+            "description": "For parcel carrier orders, this is the date the order needs to ship by.",
+            "type": "string",
+            "format": "date"
+          },
           "computed": {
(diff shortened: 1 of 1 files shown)


igregson:
> I could not remember how to update this spec in the UI of https://azurestandard.stoplight.io/docs/api-spec/public.yaml so I just modified the code manually in the spec.json [...]

No worries. Doesn't matter if you use Stoplight or not, which is the beauty of the workflow we've adopted.

igregson (changes requested):
Looks like we should also add `parcelCarrierShipByDate` to the `order` return object, no?

dwmorrison33:
> Looks like we should also add parcelCarrierShipByDate to the order return object, no?

@igregson You are correct, I have updated it.

spec.json
```diff
@@ -11095,6 +11095,11 @@
             "description": "If true, the customer has requested to hold shipment of the order until all requested stock is available.",
             "type": "boolean"
           },
+          "parcelCarrierShipByDate": {
+            "description": "For parcel carrier orders, this is the date the order needs to ship by.",
+            "type": "string",
+            "format": "date"
```
igregson, line 11101: I'm pretty sure the `format` (here and below) should be `date-time`.

dwmorrison33: It is a date, not a datetime

igregson: Hmm... I remember seeing a date with a time (in ISO format) returned by the API (while @davidmcatee-azure was demoing this for me).

dwmorrison33: Here is a reference: https://github.com/azurestandard/beehive/pull/5188#issuecomment-747064555

igregson: The value that the frontend will be setting for this ([here](https://github.com/azurestandard/website/blob/dm/parcel-carrier-ship-by/src/core/services/app_state.js#L666)) includes the time. The value comes from what the API returns as a `next-purchase-arrival` (on the `packaged-products` endpoint, though the response value here in the spec looks like it's missing this – I'll create a PR for that).

igregson: (I've created the PR for adding `next-purchase-arrival` to the `packaged-products` endpoint. Note that it's a `date-time`. See https://deploy-preview-305--azure-api-docs.netlify.app/#operation/findPackagedProduct)

dwmorrison33: @davidmcatee-azure When we discussed this last week, we agreed that the field would be a `DateField` in Django and the code was written with that requirement. Has something changed where this is supposed to be a `DateTimeField` now?

davidmcatee-azure: Since only the date matters, it makes sense to store it as a `DateField`. As far as the API is concerned, I mentioned that I wanted it in ISO ([ISO 8601](https://en.wikipedia.org/wiki/ISO_8601)) format because that would be the easiest for JavaScript to handle. These requirements are not incompatible. "2020-12-22" is valid as is "2020-12-22T16:13:04+00:00". I think keeping `"format": "date"` in the API spec would be OK. We happen to set it using a date with a time, but the time will portion of it will be dropped (this detail could be added to the description to clear up confusion).

igregson: > [...] this detail could be added to the description to clear up confusion

@dwmorrison33 Could you do this?

dwmorrison33: Yes, I updated it and pushed the code up

igregson: Perfect! Thank you :)

igregson (approved):

davidmcatee-azure (approved):

dwmorrison33:
@igregson I have merged this into master. I am gonna close this PR, but I am unsure how to deploy these updates to Spotlight.

igregson:
> I am unsure how to deploy these updates to Spotlight

There is nothing to deploy to Stoplight.

Issue #307, opened 2021-01-29
[API parity] Document PATCH on stop endpint


igregson:
Source code here: https://github.com/azurestandard/beehive/blob/master/apps/api/views/stop.py#L646-L678

PR #308, opened 2021-01-29
add truckNumber to the api spec for stop endpoint

dwmorrison33:
I added the truckNumber to the properties on the stop endpoint.

### DIFF
--- a/spec.json
+++ b/spec.json
@@ -10296,7 +10296,7 @@
           },
           "truckNumber": {
             "type": "integer",
-            "description": "The truck_number (Pickset.truck_number). Only returned if the inline query parameter includes `truckloadId` and the trip is verified."
+            "description": "The truck number of the truckload (pickset) (pickset.truck_number). Only returned if the inline query parameter includes `truckloadId` and the trip is verified."
           }
         },
         "required": [
--- a/spec.json
+++ b/spec.json
@@ -10293,6 +10293,10 @@
           "truckloadId": {
             "type": "integer",
             "description": "ID of the truckload (pickset). This will only be included if the inline query parameter includes `truckloadId` and the trip is verified."
+          },
+          "truckNumber": {
+            "type": "integer",
+            "description": "The truck_number (Pickset.truck_number). Only returned if the inline query parameter includes `truckloadId` and the trip is verified."
           }
         },
         "required": [


spec.json
```diff
@@ -10293,6 +10293,10 @@
           "truckloadId": {
             "type": "integer",
             "description": "ID of the truckload (pickset). This will only be included if the inline query parameter includes `truckloadId` and the trip is verified."
+          },
+          "truckNumber": {
+            "type": "integer",
+            "description": "The truck_number (Pickset.truck_number). Only returned if the inline query parameter includes `truckloadId` and the trip is verified."
```
igregson, line 10299: A possible improvement on the first sentence:

> The truck number of the truckload (pickset) (`pickset.truck_number`).

What do you think?

igregson:
Is it not a deviation from the existing behavior of the API's "inline" feature that each top-level property requires the property to be passed in order for it to be returned? I.e. Here, `truckNumber` is returned if `truckloadId` is passed in the inline (it doesn't require passing `truckNumber` in the inline). I'm fine with either approach, just want to make sure we're not introducing something that's breaking established patterns.

spec.json
dwmorrison33: Yes, this works. I will update it

dwmorrison33:
> Is it not a deviation from the existing behavior of the API's "inline" feature that each top-level property requires the property to be passed in order for it to be returned? I.e. Here, `truckNumber` is returned if `truckloadId` is passed in the inline (it doesn't require passing `truckNumber` in the inline). I'm fine with either approach, just want to make sure we're not introducing something that's breaking established patterns.

I see what you are saying here, but one thought is that we would never return the `truckNumber` without the `truckloadId`. So from that standpoint, I would think that it is sufficient to stay with `truckloadId`. What are your thoughts on that?

igregson:
> > Is it not a deviation from the existing behavior of the API's "inline" feature that each top-level property requires the property to be passed in order for it to be returned? I.e. Here, `truckNumber` is returned if `truckloadId` is passed in the inline (it doesn't require passing `truckNumber` in the inline). I'm fine with either approach, just want to make sure we're not introducing something that's breaking established patterns.
> 
> I see what you are saying here, but one thought is that we would never return the `truckNumber` without the `truckloadId`. So from that standpoint, I would think that it is sufficient to stay with `truckloadId`. What are your thoughts?

Yes, good point. Given that, however, maybe we make the inline property here `truckload` and have it return the [truckload object](https://github.com/azurestandard/api-spec/blob/master/spec.json#L13079-L13097)?

If we go that route, I think it would make sense to use different descriptions for that object on the trip and stop. That can be done by using the `allOf` property ([here's one place we're doing this currently](https://github.com/azurestandard/api-spec/blob/master/spec.json#L10277-L10284)).

dwmorrison33:
> Given that, however, maybe we make the inline property here truckload and have it return the truckload object?

@igregson When you say "truckload object", do you mean returned as a dict? Something like the following response example:
```
[
    {
        "estimatedDelivery": "2021-01-10T09:45:00-06:00",
        "trip": 37750,
        "drop": 794195,
        "dropId": 794195,
        "truckload": {
            "truckNumber": 2,
            "truckloadId": 48078,
        },
        "drop-shipping-fee": {
            "percent": 8.5
        },
        "timezone": "America/Chicago",
        "actualDeliveryBackdated": false,
        "id": 322264
    },
    {
        "estimatedDelivery": "2021-01-12T08:30:00-07:00",
        "finalizedDelivery": "2021-01-04T13:00:00-07:00",
        "trip": 37750,
        "truckload": {
            "truckloadId": 48076,
            "truckNumber": 1,
        },
        "pickup": 39540,
        "pickupId": 39540,
        "timezone": "America/Denver",
        "actualDeliveryBackdated": false,
        "id": 327478
    }
]
```

igregson:
> > Given that, however, maybe we make the inline property here truckload and have it return the truckload object?
> 
> @igregson When you say "truckload object", do you mean returned as a dict? Something like the following response example [...]

Yes, but following the the [truckload schema](https://github.com/azurestandard/api-spec/blob/master/spec.json#L13079-L13097). So:

```json
"truckload": {
    "truckNumber": 2,
    "id": 48078,
},
```

dwmorrison33:
@igregson Cool, I can do that.

dwmorrison33:
@igregson I know this PR is specifically for the GET api/stops endpoint, but we have logic in the PATCH api/stop endpoint that is reliant on the truckloadId in the payload, so I can update that on the backend for that too. However, I imagine you will need to send the payload like this now:
```
{
    "id":326271,
    "truckload": {
        "truckNumber": 2,
        "id": 48078,
}}
```

igregson:
> @igregson I know this PR is specifically for the GET api/stops endpoint, but we have logic in the PATCH api/stop endpoint that is reliant on the truckloadId in the payload, so I can update that on the backend too. However, I imagine you will need to send the payload like this now:
> 
> ```
> {
>     "id":326271,
>     "truckload": {
>         "truckNumber": 2,
>         "id": 48078,
> }}
> ```

It's not a problem for the frontend to send that if that that's what's needed.

Note that we don't yet have the PATCH documented for the stop endpoint (opened #307 for that yesterday).

dwmorrison33:
Closed in favor of https://github.com/azurestandard/api-spec/pull/309

PR #310, opened 2021-02-02
Dwm/add stop patch

dwmorrison33:
Resolves: https://github.com/azurestandard/api-spec/issues/307

I already added the truckloadId field to the request body even though that field is not yet on the Master branch. It soon will be with the truckload support repos.

### DIFF
--- a/spec.json
+++ b/spec.json
@@ -3432,6 +3432,26 @@
             }
           }
         }
+      },
+      "patch": {
+        "summary": "",
+        "operationId": "patch-stop-id",
+        "responses": {
+          "200": {
+            "description": "OK"
+          }
+        },
+        "description": "Partially Update a Stop",
+        "parameters": [
+          {
+            "schema": {
+              "type": "string"
+            },
+            "in": "query",
+            "name": "persist",
+            "description": "If true, calculate the offset between the trips's delivery start and this stop's estimated time. For any pre-cutoff trips associated with this stop's trip's route which cutoff later than this stop's trip, create a stop for this location (drop or pickup) if it does not already exist and update that stop's target and estimated times to have the same offset from that stops's trip's delivery start. Also create a route-stop association, if it does not already exist, for this location on this stop's trip's route and update that route-stop with the same delivery offset. If propagate is set, apply the persistance logic to the later stops on the trip as well. API users  should always pass `persist=true` if there are estimated delivery changes. For  finalized and/or actual delivery changes, passing `persist=true` will not actually  persist them but will slow down the response."
+          }
+        ]
       }
     },
     "/audit/products": {
--- a/spec.json
+++ b/spec.json
@@ -3434,24 +3434,64 @@
         }
       },
       "patch": {
-        "summary": "",
-        "operationId": "patch-stop-id",
-        "responses": {
-          "200": {
-            "description": "OK"
-          }
-        },
-        "description": "Partially Update a Stop",
+        "summary": "PATCH update for an existing stop",
+        "operationId": "partiallyUpdateStop",
+        "tags": [
+          "stop"
+        ],
+        "description": "Stop to partially update",
         "parameters": [
           {
+            "name": "id",
+            "in": "path",
+            "description": "ID of stop to update.",
+            "required": true,
             "schema": {
-              "type": "string"
-            },
+              "type": "integer",
+              "format": "int64"
+            }
+          },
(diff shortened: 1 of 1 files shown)


spec.json
```diff
@@ -10353,6 +10413,35 @@
...
+            "type": "string",
+            "format": "date-time"
+          },
+          "actualDeliveryBackdated": {
+            "type": "boolean",
+            "description": "The acutal delivery was manually set to a datetime in the past."
+          },
+          "truckloadId": {
+            "type": "integer",
+            "format": "int64"
```
igregson, line 10441: Want to add a description here? Perhaps something like: 

> Change the stop's truckload (pickset) by passing the ID of the truckload to move it to. This only works with pickups (it does not work with drops).

spec.json
```diff
@@ -10353,6 +10413,35 @@
           "actualDeliveryBackdated"
         ]
       },
+      "partiallyUpdateStop": {
+        "description": "A partial update to a trip stop or waypoint.",
+        "type": "object",
+        "properties": {
+          "estimatedDelivery": {
+            "description": "Editable until verification.",
```
igregson, line 10421: Since this is in the request body (not the response), it seems preferable for this wording to be more "action" oriented. Perhaps:

> Set or update the stop's estimated delivery datetime. This is editable until verification.

spec.json
```diff
@@ -10353,6 +10413,35 @@
...
+            "type": "string",
+            "format": "date-time"
+          },
+          "finalizedDelivery": {
+            "description": "Editable between verification and delivery.",
```
igregson, line 10426: Same as above (would be good to make this more action oriented). Perhaps:

> Set or update the stop's finalized delivery datetime. This is editable between verification and delivery.

spec.json
```diff
@@ -10353,6 +10413,35 @@
...
+            "type": "string",
+            "format": "date-time"
+          },
+          "actualDelivery": {
+            "description": "When the stop is actually reached.",
```
igregson, line 10431: Same as above (would be good to make this more action oriented). Perhaps:

> Set or update the stop's actual delivery datetime. The API only allows this before the `deliveryEnd` of the stop's trip.

igregson (changes requested):
Thanks for hoping on this! 

Left comments on ways I think we can improve the descriptions. I know with some of them you were simply taking what exists elsewhere in the spec, but even these (where they currently exist) could be updated to better match the higher level of quality we're trying to move this towards (but of course improving those is outside of the current scope).

spec.json
dwmorrison33: I am removing the the part "This only works with pickups (it does not work with drops)." as you can update a pickup or drop with the added logic. https://github.com/azurestandard/beehive/blob/58432ae704dec743e39e31c069e45217e22a94d0/apps/api/views/stop.py#L692-L697

spec.json
```diff
@@ -10418,27 +10418,28 @@
...
           "actualDeliveryBackdated": {
             "type": "boolean",
-            "description": "The acutal delivery was manually set to a datetime in the past."
+            "description": "Set or update whether the stop's actual delivery time was manually set to a datetime in the past."
           },
           "truckloadId": {
             "type": "integer",
-            "format": "int64"
+            "format": "int64",
+            "description": "Update the stop's truckload (pickset) by passing the ID of the truckload to change it to. This works with pickups or drops"
```
igregson, line 10442: Didn't know this was technically possible for drops, so it seems good to explicitly state that it works with both. How about: 

> This works regardless of whether the stop is a pickup or a drop.

spec.json
```diff
@@ -10418,27 +10418,28 @@
...
           },
           "finalizedDelivery": {
-            "description": "Editable between verification and delivery.",
+            "description": "Set or update the stop's finalized delivery datetime. This is editable between verification and delivery.",
             "type": "string",
             "format": "date-time"
           },
           "actualDelivery": {
-            "description": "When the stop is actually reached.",
+            "description": "Set or update the stop's actual delivery datetime. The API only allows this before the deliveryEnd of the stop's trip.",
```
igregson, line 10431: Since deliveryEnd is referring to a property (i.e. it is code), can you enclose it in backtics?

igregson:
Want to go ahead and update the two other instances of "actual" being misspelled (as "acutal")?

igregson (approved):

PR #314, opened 2021-03-16
Add ordered-packaged-products endpoint

davidmcatee-azure:
Implemented by https://github.com/azurestandard/beehive/pull/5330.

### DIFF
--- a/spec.json
+++ b/spec.json
@@ -7792,6 +7792,53 @@
         }
       }
     },
+    "/person/{id}/ordered-packaged-products": {
+      "get": {
+        "summary": "Returns aggregated information about packaged-products ordered by the given customer within the last two years",
+        "operationId": "listOrderedPackagedProducts",
+        "tags": [
+          "person",
+          "product"
+        ],
+        "parameters": [
+          {
+            "name": "id",
+            "in": "path",
+            "description": "ID of person whose ordered packaged products to fetch.",
+            "required": true,
+            "schema": {
+              "type": "integer",
+              "format": "int64"
+            }
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "Ordered packaged products response.",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "type": "array",
+                  "items": {
+                    "$ref": "#/components/schemas/orderedPackagedProduct"
+                  }
+                }
+              }
+            }
+          },
+          "default": {
+            "description": "Unexpected error.",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/error"
+                }
+              }
+            }
+          }
+        }
+      }
+    },
     "/session": {
       "get": {
         "summary": "Returns information about the current cookie-based session",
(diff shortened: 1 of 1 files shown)


spec.json
```diff
@@ -13207,6 +13254,43 @@
...
+            "type": "integer",
+            "format": "int64"
+          },
+          "lastOrderInvoiceDate": {
+            "description": "The invoice date of the most recent order containing the packaged-product.",
+            "type": "string",
+            "format": "date"
+          },
+          "lastOrderId": {
+            "description": "The id of the most recent order containing the packaged-product.",
```
igregson, line 13282: Should "id" be uppercased (like you did on line 13267)?

spec.json
```diff
@@ -7792,6 +7792,53 @@
...
+        "operationId": "listOrderedPackagedProducts",
+        "tags": [
+          "person",
+          "product"
+        ],
+        "parameters": [
+          {
+            "name": "id",
+            "in": "path",
+            "description": "ID of person whose ordered packaged products to fetch.",
```
igregson, line 7807: Since descriptions should be formatted as sentences (according to [our style guide](https://github.com/azurestandard/api-spec#style-guide)), I would start this with "The" and do the same for the other descriptions in this PR where it makes sense.

igregson (changes requested):

spec.json
```diff
@@ -7792,6 +7792,53 @@
...
+        "operationId": "listOrderedPackagedProducts",
+        "tags": [
+          "person",
+          "product"
+        ],
+        "parameters": [
+          {
+            "name": "id",
+            "in": "path",
+            "description": "The ID of the person whose ordered packaged products to fetch.",
```
igregson, line 7807: This sentence seems to read better if "to fetch." is "that should be fetched." That would make this:

> The ID of the person whose ordered packaged products should be fetched.

What do you think?

spec.json
```diff
@@ -13207,6 +13254,43 @@
           "id",
           "truckNumber"
         ]
+      },
+      "orderedPackagedProduct": {
+        "description": "Aggregated information about packaged-products ordered by a customer within the last two years.",
+        "type": "object",
+        "properties": {
+          "code": {
+            "description": "The packaged product code that was ordered.",
```
igregson, line 13263: It seems that maybe we should hyphenate "packaged product" everywhere, no?

spec.json
davidmcatee-azure: So, no "that" then?

igregson: ah, yes... missed that. No "that".

spec.json
davidmcatee-azure: It looks like within summaries and descriptions of the spec, there are more unhyphenated "packaged product" occurrences than hyphenated "packaged-product" occurrences. Perhaps it would rather make more sense for me to stick with the unhyphenated versions then?

igregson: Ah, good thinking to check the existing usage (since consistency with this type of thing is important). I'm fine going either direction on this.

davidmcatee-azure: I went with removing the hyphen. There are a few uses of hyphenated versions elsewhere in the file but I won't change those in this scope.

igregson (approved):

PR #319, opened 2021-05-03
Add finalized delivery state param to stop

igregson:
For https://github.com/azurestandard/beehive/issues/5444.

In discussion with Tom and Thomas we were thinking to name the property `finalizedState`. While creating this PR, however, it became clear to me that a more descriptive name is `finalizedDeliveryState` (further reasoning: The date-time property is named `finalizedDelivery`, and this property essentially describes the state of that property, so "extending" its name seems to make good sense (i.e. `finalizedDelivery + [state]`)). Of course feel free to share any reasons we may not want to go with `finalizedDeliveryState`.

### DIFF
--- a/spec.json
+++ b/spec.json
@@ -10424,6 +10424,20 @@
                 "$ref": "#/components/schemas/truckload"
               }
             ]
+          },
+          "finalizedDeliveryState": {
+            "oneOf": [
+              {
+                "enum": [
+                  "confirmed",
+                  "driverWillCall",
+                  "toBeDetermined",
+                  "unknownDelay"
+                ],
+                "description": "The state of the stop's finalized delivery. This is always returned if a `finalizedDelivery` value is returned."
+              }
+            ],
+            "type": "string"
           }
         },
         "required": [
--- a/spec.json
+++ b/spec.json
@@ -10425,16 +10425,15 @@
               }
             ]
           },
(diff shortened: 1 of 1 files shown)


igregson:
@tompetersjr and @tommckennon, wasn't sure whose review to request on this so requested one from both of you.

spec.json
```diff
@@ -10424,6 +10424,20 @@
                 "$ref": "#/components/schemas/truckload"
               }
             ]
+          },
+          "finalizedDeliveryState": {
+            "oneOf": [
```
tommckennon, line 10429: Wondering why the oneOf property is defined within finalizedDeliveryState. I didn't see it defined for other Enum types, so wondering if it effects how to return the string value?

igregson: `oneOf` is a [keyword in the OpenAPI 3 Specification](https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/).

There are other instances in the spec file where we use `oneOf` (for reference).

tommckennon: Thanks

igregson:
In conversation with Thomas it's become clear that a better name for this property is `finalizedDeliveryFlag`. We also realized that there's no value added in including `confirmed` (and that it would perhaps even introduce confusion). I'll update this PR accordingly.

tommckennon (approved):

PR #321, opened 2021-06-24
Picksets-Trucks api initial defintions

tommckennon:


spec.json
```diff
@@ -7505,6 +7669,18 @@
               "type": "integer",
               "format": "int32"
             }
+          },
```
tommckennon, line 7672: Note: This is a little deceiving in GitHub, but this is under the /people endpoint definition.

spec.json
```diff
@@ -7505,6 +7669,18 @@
...
+          {
+            "name": "customerTypes",
```
davidmcatee-azure, line 7674: It would be simpler at first to make this filter by a single `customerType`. When allowing multiple, there are two ambiguities that are introduced. One is whether the multiple filters function as an AND vs. an OR (it could be either since this is a many-to-many relationship). The other is how the multiple values are serialized/deserialized. Both ambiguities would need to be clearly defined and it doesn't make sense to go through all that when all we currently need is to filter by one type, i.e. "truck-driver".

davidmcatee-azure (changes requested):

spec.json
```diff
@@ -3504,6 +3504,170 @@
         }
       }
     },
+    "/trucks": {
```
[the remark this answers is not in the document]
tommckennon: Hey David. I want to make sure of this. Perhaps trucks is the wording to use here. I'm seeing that the "truckloads" synonym is already being used in the api for the internal website here: https://github.com/azurestandard/api-spec/blob/28dae5fce84c35f9b1d3178917b5cdac80f439c7/spec.json#L12993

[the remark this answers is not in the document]
tommckennon: Tom mentioned here https://github.com/azurestandard/beehive/pull/5603#discussion_r658075697 we were planning to have the truck drivers be apart of the already defined "truckloads" and want to instead make them apart of the new trucks api if I'm reading that correctly. But I may be misinterpreting, so seeking confirmation this should not be "trucks" before changing to "truckloads".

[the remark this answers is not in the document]
davidmcatee-azure: This "truckload" is the same thing as what you are implementing; "truckloads" is a synonym for "picksets". Currently the only way truckloads are exposed to the frontend is if you inline them on a /trips request using `inline=truckloads` or on a /stops request using `inline=truckload`. The new endpoint `/truckloads` is the first-class Plural endpoint for the same resource. As I mentioned in another spot, we don't actually need this Plural endpoint right now since we are inlining the truckloads in both our queries to /trips and /stops (an alternative implementation without using the inline parameter could use separate queries to the Plural truckloads endpoint filtering for a particular trip or stop, but it was decided in the initial implementation of truckload support to take a shortcut and only implement the inline support rather than make the first-class endpoint for truckloads). What we do need is a `put` endpoint for the Singular `/truckload/{id}` due to the fact that we need to be able to update the driver on a truckload from the frontend.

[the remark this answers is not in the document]
davidmcatee-azure: It is the Singular `put` that is necessitating the creating of the first-class /truckloads endpoint. But it is the same thing as the "truckloads" that already exists in the spec.

spec.json
```diff
@@ -3504,6 +3504,170 @@
...
+            "name": "truckDriver",
+            "in": "query",
+            "description": "Only return trucks with this truck driver id.",
+            "required": false,
+            "style": "form",
+            "schema": {
+              "type": "integer",
+              "format": "int32"
+            }
+          }
```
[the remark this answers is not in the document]
tommckennon: That makes sense. Will remove those. Actually, after thinking about it, I think the first filter would be a filter on route_trip to get the pick sets of the route trip, but not sure if it's needed based on what's already being pulled into logistics. Will add route_trip if making that query makes sense somewhere in the workflow.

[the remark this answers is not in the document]
davidmcatee-azure: > but not sure if it's needed based on what's already being pulled into logistics

Right, this relates to my explanation of why it should be "truckloads". The reason it's not needed is because we're already inlining the truckloads in our /trips request.

spec.json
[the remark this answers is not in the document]
tommckennon: I see. Understood. It makes sense to match it to how it's already defined in this api under trips, which makes "truckloads" the correct term. Thanks. Changing.

spec.json
tommckennon: There's no question we will be adding more, so I think we'd want this to stay. I've handled many to many relationships by always passing an array of values to filter on with an array length of 1 for one value to filter on, since the relationship is not equal to a singular value (always an array of values).

davidmcatee-azure: We may be adding more customer types, but I am questioning whether we have a use case for filtering on more than one. If not, then I would start with implementing filtering by a single `customerType`. If we have a use case arise later to filter on multiple, then we can update accordingly. The parameter name should be `customerType` either way (not the plural `customerTypes`). In other places where we have query parameters that accept multiple values, the singular form is still used.

tommckennon: I follow. That's the one facet of the api I've done differently on my end semantically, but the reasoning and approach you mention works for me...changing.

tommckennon:
Thank you for covering all the detailed fixes for these api additions. It puts me in position to put forth a PR for the corresponding API provider updates.

spec.json
```diff
@@ -10081,6 +10234,37 @@
...
+          },
+          "pick-date": {
+            "description": "Date-time that orders for this truck should be picked.",
+            "type": "string",
+            "format": "date-time"
+          }
+        },
+        "required": [
+          "id",
+          "truckDriverId"
```
davidmcatee-azure, line 10265: If `"truckDriverId"` is required, then you could not do a PUT with an empty `truckDriverId` to remove the truck driver from the truckload. In other words, once a truck driver is assigned, you can only change the truck driver, you cannot remove. I don't think we had clear requirements on this aspect. It is a bit restrictive on the user to not allow removal (what if they accidentally assigned a truck driver prematurely, and wanted to undo it?) However, since we don't have clear requirements there, I would be OK with requiring `truckDriverId`, but then your backend implementation should match. I will comment accordingly.

tommckennon: Understood. Did so on the updates to https://github.com/azurestandard/beehive/pull/5603

spec.json
```diff
@@ -3504,6 +3504,147 @@
         }
       }
     },
+    "/truckloads": {
+      "get": {
+        "summary": "Returns all truckloads (picksets) from the system that the user has access to",
+        "description": "Truckloads (picksets) are ordered by truck number.",
```
davidmcatee-azure, line 3510: It looks like your implementation uses the default `PickSet` ordering of `"route_trip__pull", "sequence", "truck_number"`. I don't think this description adequately reflects that, but I'm also not sure how to translate that into English because I don't know what "sequence" is. This would at least be closer:

> "Truckloads (picksets) are ordered by the associated trip's pick date and then truck number."


>

spec.json
```diff
@@ -10081,6 +10234,37 @@
...
+          },
+          "truckDriverId": {
+            "description": "Truck Driver id of truck.",
+            "type": "int64"
+          },
+          "pick-date": {
+            "description": "Date-time that orders for this truck should be picked.",
+            "type": "string",
+            "format": "date-time"
+          }
```
davidmcatee-azure, line 10261: You can remove `route-trip`, `truck-number`, and `pick-date` since we are not currently allowing those to be updated via the API.

davidmcatee-azure (approved):

PR #322, opened 2021-07-13
Add allergens to product properties

tommckennon:


tommckennon:
@davidmcatee-azure , shall open another PR for the product api spec fixes for the product api spec issues I'm seeing, or add them here?

davidmcatee-azure:
A separate PR.

tommckennon:
Updated: As noted, we only need to return the product's allergen names for display purposes on the website.

spec.json
```diff
@@ -10689,6 +10689,14 @@
...
             "type": "string"
           },
+          "allergens": {
+            "description": "The names of the Allergens used for labeling of this product.",
+            "type": "array",
+            "items": {
+              "description": "The Allergen name.",
+              "type": "string"
+            }
+          },
```
davidmcatee-azure, line 10699: I can't tell if there is any rhyme or reason to the order that properties are listed here. In the absence of such, I would rather have this go at the end than inserted here near the top (unless you have a reason for putting it here?).

tommckennon: The only rhyme and reason that I have on my mind is to have the model field properties placed above the model class' custom property attributes. That would mean placing it above the slug property rather than at the bottom. But it's not necessary, especially when going to using ViewSets on the backend which would take care of that order of properties returned for us. Thus, I'll strike that idea of model field properties  at the top.

tommckennon: After thinking about it some more while fixing the get Product properties (https://github.com/azurestandard/api-spec/pull/323), I placed the allergens back to alongside the baseline properties. I found easeir to track the returned properties by having the baseline properties and then the inline properties defined, as they are specified in the backend.

davidmcatee-azure (approved):

spec.json
davidmcatee-azure: Can you fix the merge conflict now that I've merged the other PR?

tommckennon: Yes, was expecting that.

tommckennon: Set

spec.json
```diff
@@ -10697,6 +10697,9 @@
             "description": "If true, this product will be published to search engines even when it is inactive (out for season, out long term, or discontinued).",
             "type": "boolean"
           },
+          "inactiveDescription": {
+            "description": "HTML description applicable only when the product is inactive (out for season, out long term, or discontinued). It should contain specific information pertinent to the inactive status of the product (e.g. details about why it is inactive and what similar products are available instead). It should not duplicate information in short-description or description.",
+            "type": "string"
```
davidmcatee-azure, line 10702: It looks like these three lines need to be removed.

tommckennon: Thank you. A merge conflict that needed fixing. Set

davidmcatee-azure (approved):

PR #323, opened 2021-07-14
Get Product Property Fixes

### LINKED ISSUE #306: [API parity] Update product response object to what API actually returns
The response object for the `product` endpoint (https://azure-api-docs.netlify.app/#operation/findProductById) is not accurate. We should update it to what the API actually returns.

tommckennon:
Resolves #306

### DIFF
--- a/spec.json
+++ b/spec.json
@@ -10723,7 +10723,7 @@
             "type": "string"
           },
           "packaging": {
-            "description": "Packaged versions of this product.",
+            "description": "Packaged versions of this product. This will only be included if the `inline` query parameter includes `packaging`.",
             "type": "array",
             "items": {
               "type": "integer",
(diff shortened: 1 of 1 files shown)


tommckennon:
Order fixed as well as to match how these properties are returned from the api.

tommckennon:
I also ordered these properties to match how they are specified on the backend. I think it makes it easier to follow what properties are defined for the api whether baseline or inline.

spec.json
```diff
@@ -10686,60 +10686,85 @@
             "format": "int64"
           },
           "name": {
-            "description": "Sentence-length product name.",
+            "description": "Sentence-length product name (either its given name or marketing name).",
```
davidmcatee-azure, line 10689: From the API consumer's perspective, the distinction between given name or marketing name does not matter. I think this edit should be reverted.

spec.json
```diff
@@ -10686,60 +10686,85 @@
...
           },
           "treatAsActive": {
             "description": "If true, this product will be published to search engines even when it is inactive (out for season, out long term, or discontinued).",
             "type": "boolean"
           },
-          "inactiveDescription": {
-            "description": "HTML description applicable only when the product is inactive (out for season, out long term, or discontinued). It should contain specific information pertinent to the inactive status of the product (e.g. details about why it is inactive and what similar products are available instead). It should not duplicate information in short-description or description.",
+          "storageClimate": {
+            "description": "The description of the product's storage climate.",
             "type": "string"
```
davidmcatee-azure, line 10702: It seems to me this can be kept as `"$ref": "#/components/schemas/storage"`.

spec.json
```diff
@@ -10686,60 +10686,85 @@
...
           },
-          "short-description": {
-            "description": "Paragraph-length, stand-alone, plain-text product description.",
-            "type": "string"
+          "isShippableUps": {
+            "description": "If true, this product will can be shipped by UPS.",
```
davidmcatee-azure, line 10705: "If true, this product can be shipped by parcel carrier."

spec.json
```diff
@@ -10686,60 +10686,85 @@
...
+          "seoName": {
+            "description": "An SEO-optimized name. A lengthened version of the full name that includes additional keywords or rephrasing, suitable for use in a title tag. When this is unset, the name is a reasonable fallback.",
             "type": "string"
           },
+          "packaging": {
+            "description": "Packaged versions of this product. This will only be included if the `inline` query parameter includes `packaging`.",
+            "type": "array",
+            "items": {
+              "type": "integer",
+              "format": "int64"
```
davidmcatee-azure, line 10730: Should be:
`"$ref": "#/components/schemas/packagedProduct"`

spec.json
```diff
@@ -10686,60 +10686,85 @@
...
           "brand": {
-            "description": "The brand associated with this product.",
+            "description": "The brand associated with this product. This will only be included if the `inline` query parameter includes `brand`.",
             "type": "integer",
             "format": "int64"
           },
-          "directions": {
-            "description": "Product use directions (e.g., cooking suggestions).",
+          "keywords": {
+            "description": "Comma-separated list of keywords that will be used in Google web search for this specific product. This will only be included if the `inline` query parameter includes `keywords`.",
```
davidmcatee-azure, line 10739: Suggest simplifying to:
"Comma-separated list of keywords for this product. This will only be included if the `inline` query parameter includes `keywords`."

spec.json
```diff
@@ -10686,60 +10686,85 @@
...
             "type": "string"
           },
-          "favorites": {
-            "description": "Number of people who've marked this product as a favorite.",
-            "type": "integer",
-            "format": "int32"
+          "description": {
+            "description": "Long-form, HTML product description. When short-description is set, this value is for additional information which did not fit into short-description. Displaying both short-description and description is not redundant. Displaying only description may be confusing when short-description is available. This will only be included if the `inline` query parameter includes `description`.",
```
davidmcatee-azure, line 10743: Replace all instances of "short-description" with `shortDescription`.
For example: " ... When `shortDescription` is set ... "

spec.json
```diff
@@ -10760,15 +10785,52 @@
...
+                },
+                "buyer": {
+                  "description": "Customer ID (number) for the vendor's buyer.",
+                  "type": "int64"
+                },
+                "indie": {
+                  "description": "If true, the vendor is a privately held company.",
+                  "type": "boolean"
+                }
+              }
```
davidmcatee-azure, line 10829: Should reference existing vendor schema:
```
                  "items": {
                    "$ref": "#/components/schemas/vendor"
                  }
```

davidmcatee-azure (changes requested):

spec.json
tommckennon: This one (storageClimate) is returning just a string: https://github.com/azurestandard/beehive/blob/ada64c39d27cebc87ec1d99124c98934a252cc28/apps/api/views/product.py#L51

davidmcatee-azure: `"$ref": "#/components/schemas/storage"` refers to this:

https://github.com/azurestandard/api-spec/blob/tm/get-product-fixes/spec.json#L10482-L10491

... which is a string. It can only have 4 values: "dry", "chilled", "frozen", or "greenhouse".

spec.json
```diff
@@ -10726,29 +10726,28 @@
...
+            "description": "Comma-separated list of keywords for this product. This will only be included if the `inline` query parameter includes `keywords`.",
             "type": "string"
           },
           "description": {
-            "description": "Long-form, HTML product description. When short-description is set, this value is for additional information which did not fit into short-description. Displaying both short-description and description is not redundant. Displaying only description may be confusing when short-description is available. This will only be included if the `inline` query parameter includes `description`.",
+            "description": "Long-form, HTML product description. When shortDescription is set, this value is for additional information which did not fit into shortDescription. Displaying both shortDescription and description is not redundant. Displaying only description may be confusing when shortDescription is available. This will only be included if the `inline` query parameter includes `description`.",
```
davidmcatee-azure, line 10742: Each instance of `shortDescription` should be enclosed in backticks.

spec.json
tommckennon: I see. Ready for review.

spec.json
```diff
@@ -10699,7 +10699,10 @@
           },
           "storageClimate": {
             "description": "The description of the product's storage climate.",
-            "type": "string"
+            "type": "string",
+            "items": {
+              "$ref": "#/components/schemas/storage"
+            }
```
davidmcatee-azure: This should be:
```
          "storageClimate": {
            "$ref": "#/components/schemas/storage"
          },
```

spec.json
```diff
@@ -10699,10 +10699,7 @@
           },
           "storageClimate": {
             "description": "The description of the product's storage climate.",
-            "type": "string",
-            "items": {
-              "$ref": "#/components/schemas/storage"
-            }
+            "$ref": "#/components/schemas/storage"
```
davidmcatee-azure, line 10702: You can also remove the description since the ref has its own description and it is the one that is displayed in redoc.

davidmcatee-azure (approved):

PR #332, opened 2021-11-22
Add id to promoCode response object

igregson:
This is needed for the frontend to use some of the promo-code endpoints.

For context, see: 

- https://github.com/azurestandard/beehive/pull/5884#issuecomment-974141741
- https://github.com/azurestandard/beehive/pull/5884#issuecomment-974258070


tommckennon (approved):
Approved

PR #333, opened 2021-12-13
Add piece meta links to product api.

tommckennon:


tommckennon:
@davidmcatee-azure , I figured I would give this new inline product property a more descriptive name (pieceMetaLinks) in case we want to add any other product links in the future, and the front-end will know exactly what type of links they are. If no, just name it links?

tommckennon:
@davidmcatee-azure , I added the inline rule to this new property's description. I noticed its left out of the allergens description. Probably should add that to the allergens description, correct?

davidmcatee-azure:
> @davidmcatee-azure , I figured I would give this new inline product property a more descriptive name (pieceMetaLinks) in case we want to add any other product links in the future, and the front-end will know exactly what type of links they are. If no, just name it links?

I would just name it links. As far as the API is concerned, there is no such thing as a piece meta. It is abstracted and called a "product", so "pieceMetaLinks" is no more descriptive than "links" since it is on the product JSON.

davidmcatee-azure:
> Probably should add that to the allergens description, correct?

Correct.

tommckennon:
Thanks @davidmcatee-azure for answering those questions. You can provide review these spec changes when you're free.

spec.json
```diff
@@ -11122,6 +11122,14 @@
               "$ref": "#/components/schemas/vendor"
             }
           },
+          "links": {
+            "description": "The links of this product. This will only be included if the `inline` query parameter includes `links`.",
+            "type": "array",
+            "items": {
+              "url": "string",
+              "title": "string"
+            }
```
davidmcatee-azure, line 11131: This doesn't appear to be the correct format. Redoc shows it like this:

https://deploy-preview-333--azure-api-docs.netlify.app/#operation/findProducts

image

I think it should say "array of object" and then be expandable to show the properties.

tommckennon: Right. It needs to be an array of `Links`. I added a link object to the schemas section for that purpose of having link type to reference here now. Hopefully its an appropriate spot in the schemas section based on what type of object it is.

tommckennon: Pushed the `Link` schema object with required field of url since title is optional.

spec.json
```diff
@@ -13255,6 +13262,28 @@
...
         ]
       },
+      "link": {
+        "description": "A URL link.",
+        "type": "object",
+        "properties": {
+          "id": {
+            "type": "integer",
+            "format": "int64"
+          },
```
davidmcatee-azure: I don't think including the `id` is necessary since there is nothing that can be done with it.

tommckennon: Yup. Sorry. Just removed.

davidmcatee-azure (approved):

PR #335, opened 2022-02-15
Add `replace` query param to promo code POST

igregson:
Updates the spec for changes made here: https://github.com/azurestandard/beehive/pull/5884#issuecomment-1039707758

### DIFF
--- a/spec.json
+++ b/spec.json
@@ -6356,7 +6356,18 @@
           },
           "description": "Order Promo Code to add.",
           "required": true
-        }
+        },
+        "description": "Adds a promo code to an order. Can also be used to replace an order's existing promo code.",
+        "parameters": [
+          {
+            "schema": {
+              "type": "boolean"
+            },
+            "in": "query",
+            "name": "replace",
+            "description": "Used for replacing the order's promo code with the one passed in the request."
+          }
+        ]
       }
     },
     "/order-promo-code/{id}": {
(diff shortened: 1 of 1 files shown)


spec.json
```diff
@@ -6356,7 +6356,18 @@
           },
           "description": "Order Promo Code to add.",
           "required": true
-        }
+        },
+        "description": "Adds a promo code to an order. Can also be used to replace an order's existing promo code.",
```
tommckennon, line 6360: Why is this parent description property here? It seems like it would just be needed as it is defined within the "replace" parameter's description (i.e. it seems it doesn't need it described here too).

spec.json
```diff
@@ -6356,7 +6356,18 @@
...
+        "parameters": [
+          {
+            "schema": {
```
tommckennon, line 6363: For consistency, I always see the `name` property at top and the `schema` property at the bottom of each parameter.

spec.json
igregson: Here's how these are used by Redoc:

CleanShot 2022-02-16 at 15 28 39@2x

(https://deploy-preview-335--azure-api-docs.netlify.app/#operation/addOrderPromoCode)

spec.json
igregson: I create this PR via Stoplight Studio (https://github.com/azurestandard/api-spec#stoplight-studio), I didn't manually create any of the sorting or anything (so the way the items are sorted was done by Stoplight Studio). When we adopted our workflow for utilizing Stoplight Studio a big part of the motivation was to prevent having to work with raw JSON. Given that, I'd prefer to keep the ordering as it creates it, unless there's a strong argument against this.

tommckennon (approved):

davidmcatee-azure (approved):

